PR #12, opened 2025-09-25
bugfix: Fix uptime calculation in FreeRTOS port

buckleypaul:
Resolve issue in FreeRTOS port where quantization issues could occur with timer tick values that are not even divisors of 1000.

For example, if a tick rate is 300Hz, due to the integer division to ms_per_tick would be calculated to be 3ms (vs 3.33..3ms) which would cause the later multiplication of ticks * ms_per_tick to result in an invalid ms update.


ceolin:
Comment about parenthesis and please break the commit messages to 80 columns.

buckleypaul:
> Comment about parenthesis and please break the commit messages to 80 columns.

You saw I updated the parenthesis already yeah? Or did I mess up the push?

ceolin:
> > Comment about parenthesis and please break the commit messages to 80 columns.
> 
> You saw I updated the parenthesis already yeah? Or did I mess up the push?

part of .. my suggestion is to make the precedence explicit, this helps to avoid bugs specially with rounding / casts .

`(xTaskGetTickCount() * 1000) / configTICK_RATE_HZ`

`xTaskGetTickCount() * (1000 / configTICK_RATE_HZ)`

ceolin (approved):
Thanks !!

PR #41, opened 2025-11-12
RFC: ble: utils: Add some utils when advertising only Hubble data

ceolin:
Add some helpers to facilitate when advertising data only using Hubble service.

- The same utils used across different ports (Zephyr, TI, esp-idf).


include/hubble/ble/utils.h
```diff
@@ -0,0 +1,158 @@
...
+ * @note The output format depends on the platform:
+ *       - Zephyr: Returns number of bt_data structures (typically 2)
+ *       - Other platforms: Returns total byte length of advertisement packet
+ **/
+
+#ifdef __ZEPHYR__
+
+#include <zephyr/bluetooth/bluetooth.h>
+
+#define HUBBLE_BLE_UTILS_BUFFER_SIZE 25
```
HongNguyen635: I'm wondering if I'd be better if there are headers common to both zephyr and other platforms. It might be confusing when on zephyr the utils buffer is 25 and others are 31. Maybe something like this:

```
#define BLE_ADV_LEN 31
#define HUBBLE_BLE_ADV_HEADER_SIZE 6

#ifdef __ZEPHYR__
#define HUBBLE_BLE_UTILS_BUFFER_SIZE (BLE_ADV_LEN - HUBBLE_BLE_ADV_HEADER_SIZE)
#endif
```
something like that.

samples/freertos/ti/ble-beacon/src/hubble_ble_adv.c
```diff
@@ -136,20 +126,14 @@ bStatus_t hubble_ble_adv_start(void)
...
 	if (_adv_timer_setup() == FAILURE) {
 		return (FAILURE);
 	}
 
-	len = BLE_ADV_LEN - HUBBLE_BLE_ADV_HEADER_SIZE;
-	if (hubble_ble_advertise_get(
-		    NULL, 0, &advData[HUBBLE_BLE_ADV_HEADER_SIZE], &len) != 0) {
-		return (FAILURE);
-	}
+	HUBBLE_BLE_ADV_DATA_SET(advData, NULL, 0, &len);
```
HongNguyen635: same here, no return status check?

include/hubble/ble/utils.h
```diff
@@ -0,0 +1,158 @@
...
+		_name[0].type = BT_DATA_UUID16_ALL;                            \
+		_name[0].data = (uint8_t *)&_name##_uuid;                      \
+		*(_out_len) = HUBBLE_BLE_UTILS_BUFFER_SIZE;                    \
+		_name##_err = hubble_ble_advertise_get(                        \
+			_data, _len, _name##_adv_buf, _out_len);               \
+		if (_name##_err == 0) {                                        \
+			_name[1].data_len = *(_out_len);                       \
+			_name[1].type = BT_DATA_SVC_DATA16;                    \
+			_name[1].data = _name##_adv_buf;                       \
+			*(_out_len) = 2;                                       \
```
HongNguyen635: Everything looks good to me except this line. What happened if the bt_entries is more than 2 (like the current zephyr ble_beacon sample)?

include/hubble/ble/utils.h
```diff
@@ -0,0 +1,158 @@
...
+ */
+#define HUBBLE_BLE_ADV_DEFINE(_name)                                           \
+	static uint8_t _name[HUBBLE_BLE_UTILS_BUFFER_SIZE] = {                 \
+		0x03, 0x03, 0xA6, 0xFC, 0x01, 0x16}
+
+#define HUBBLE_BLE_ADV_DATA_SET(_name, _data, _len, _out_len)                  \
+	({                                                                     \
+		int _name##_err;                                               \
+		memset(&_name[6], 0, HUBBLE_BLE_UTILS_BUFFER_SIZE - 6);        \
+		*(_out_len) = HUBBLE_BLE_UTILS_BUFFER_SIZE - 6;                \
```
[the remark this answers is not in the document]
ceolin: So ... the problem is that you start to leak symbols that are irrelevant for the application and this is a public header. We need to minimize exported symbols that are not usable by the application. I rather have a comment explaining it than a symbol that is not even used in all ports ...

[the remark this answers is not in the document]
HongNguyen635: ah I forgot about the part that this is a utility header. Yeah I guess an explanation comment would suffice then.

include/hubble/ble/utils.h
HongNguyen635: ok that makes sense. This probably needs to be updated/mentioned in the docs and README as well

ceolin:
Stale PR. Closing it.

PR #153, opened 2026-03-30
Update README.md

apoorvanori:
Updated example Espressif board from ESP32-C3 to C6


ceolin:
@danielr-hubblenetwork FYI ^

ceolin:
Stale PR, closing it but feel free to to re-open in case of new changes.

PR #159, opened 2026-04-01
github: actions: Remove bsim action

ceolin:
- Remove bsim action (bsim was not been used, only wasting resources)
- Build (and) run tests instead of only build samples
- Rename action to something simpler


.github/workflows/zephyr-tests.yaml
buckleypaul, line 1: I would prefer this to be zephyr-unit-tests.yml instead of the generic “tests” if you’re renaming it anyways.

HongNguyen635 (dismissed):
I agree with Paul's comment above, otherwise looks good to me.

.github/workflows/zephyr-tests.yaml
ceolin: It makes sense to me. I will chage it.

HongNguyen635 (approved):

buckleypaul (approved):

PR #162, opened 2026-04-02
docs: add configuration reference page

buckleypaul:
Users had no single place to find all SDK configuration options, their defaults, valid ranges, and platform availability. Config info was scattered across Kconfig help text and quickstart guides.

Add docs/configuration/index.rst covering all 19 CONFIG_HUBBLE_* options organized by category with platform-specific guidance, common examples, and a cross-platform availability table.

### DIFF
--- a/docs/configuration/index.rst
+++ b/docs/configuration/index.rst
@@ -1,0 +1,498 @@
+.. _hubble_configuration:
+
+Configuration Reference
+#######################
+
+The Hubble Network SDK is configured entirely at compile time. There is no
+runtime configuration API. All options are set before building and compiled
+into the firmware binary.
+
+The configuration mechanism depends on the platform:
+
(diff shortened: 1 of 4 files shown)


HongNguyen635 (commented):
This is nice to summary all the kconfig, but I'm just wondering if this is maintainable. These things changes all the time and when we're adding more config options this will get out of hand pretty quick I think.

For zephyr and esp-idf, use `menuconfig` will shows all the config related to Hubble Network lib but freertos doesn't have any equivalent.

I'm just gonna wait for @ceolin to input, otherwise this looks good to me for now.

ceolin (changes requested):
Generate these options from code. There is a way to bridge Kconfig options to Sphinx, this is not maintainable.

PR #163, opened 2026-04-02
feat: hard-code EID pool size to 128

buckleypaul:
Remove `CONFIG_HUBBLE_EID_POOL_SIZE` as a configurable option since only
a pool size of 128 is currently supported by the backend. Hard-code it
as a `#define` in `src/hubble.c` and remove from Kconfig, FreeRTOS config,
and test configurations.


tests/zephyr/unit/ble-advertise/src/main.c
```diff
@@ -374,7 +374,7 @@ ZTEST_SUITE(ble_advertise_format_test, NULL, ble_advertise_format_test_setup,
 
 #define TEST_COUNTER_ROTATION_MS                                               \
 	((uint64_t)CONFIG_HUBBLE_EID_ROTATION_PERIOD_SEC * 1000ULL)
-#define TEST_COUNTER_POOL_SIZE CONFIG_HUBBLE_EID_POOL_SIZE
+#define TEST_COUNTER_POOL_SIZE HUBBLE_EID_POOL_SIZE
```
ceolin: Isn't this symbol defined in the `C` file now ?

HongNguyen635 (approved):

PR #164, opened 2026-04-02
crypto: derive MbedTLS cipher type from CONFIG_HUBBLE_KEY_SIZE

buckleypaul:
The MbedTLS backend selected its AES cipher type using Kconfig boolean macros (CONFIG_HUBBLE_NETWORK_KEY_256 / CONFIG_HUBBLE_NETWORK_KEY_128). These macros are only defined by Zephyr and ESP-IDF Kconfig — the FreeRTOS port configures key size solely through the integer CONFIG_HUBBLE_KEY_SIZE, so building FreeRTOS with the MbedTLS backend would fail at compile time with "Invalid Hubble Key size".

Derive CIPHER_TYPE from CONFIG_HUBBLE_KEY_SIZE instead, matching how the PSA backend already works. The Kconfig boolean choice symbols remain unchanged — they continue to derive the integer in Zephyr and ESP-IDF builds as before.

### DIFF
--- a/src/crypto/mbedtls.c
+++ b/src/crypto/mbedtls.c
@@ -12,12 +12,12 @@
 
 #define _STREAM_BLOCK_LEN 16
 
-#if defined(CONFIG_HUBBLE_NETWORK_KEY_256)
+#if CONFIG_HUBBLE_KEY_SIZE == 32
 #define CIPHER_TYPE MBEDTLS_CIPHER_AES_256_ECB
-#elif defined(CONFIG_HUBBLE_NETWORK_KEY_128)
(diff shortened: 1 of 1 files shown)


ceolin (changes requested):
It is better to have FreeRTOS config.h using `CONFIG_HUBBLE_NETWORK_KEY_256` and `CONFIG_HUBBLE_NETWORK_KEY_128` instead. It matches with other two and gives direct control over the length.

This option is not intended to be directly controlled and ... if the user sets a different value and uses PSA you will get only runtime issue

buckleypaul:
Closing this in favor of https://github.com/HubbleNetwork/hubble-device-sdk/pull/169

PR #165, opened 2026-04-02
api: rename EID mode to counter source to match BE

buckleypaul:
The previous names "time-based" and "counter-based" did not match the backend calls which could cause confusion. Also, namespace the config values to COUNTER_SOURCE to match the backend (from EID).

- HUBBLE_EID_MODE -> HUBBLE_COUNTER_SOURCE
- HUBBLE_EID_TIME_BASED -> HUBBLE_COUNTER_SOURCE_UNIX_TIME
- HUBBLE_EID_COUNTER_BASED -> HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME
- hubble_eid_counter_get() -> hubble_counter_get()


ceolin (dismissed):
NO for Epoch time based is not the most accurate, we are using unix time.

```
Definition of "Epoch": Strictly speaking, an "epoch" is just a starting point. While the "Unix Epoch" is 1970, other systems have different epochs (e.g., GPS uses 1980, and Excel uses 1900). Therefore, "Unix time" is a more precise term for the specific 1970-based standard.
```

HongNguyen635:
I feel like `HUBBLE_EID_XXX_BASED` makes more sense (for beginners like me) because when I read it I know it relates to EID without understanding about crypto or how that is generated. When reading counter, my impression is it related to the seq counter (since we have seq counter here). But that could just be me.

buckleypaul:
> I feel like `HUBBLE_EID_XXX_BASED` makes more sense (for beginners like me) because when I read it I know it relates to EID without understanding about crypto or how that is generated. When reading counter, my impression is it related to the seq counter (since we have seq counter here). But that could just be me.

I can see that, but we've already released the counter_source nomenclature to the public with the backend here: https://hubble.com/docs/api-specification/register-new-devices#set-counter-source-for-eid-rotation

Documenting it as the "counter source for EID rotation" works for the embedded SDK as well. I want the key/value to be the same between backend and embedded, and it's sort of a wash which is better IMO.

HongNguyen635 (approved):
Change like this is expensive, I hope we don't miss any.

PR #166, opened 2026-04-03
docs: auto-generate configuration reference from Kconfig

buckleypaul:
Add a Sphinx extension (kconfig_autodoc) that parses port/zephyr/Kconfig using kconfiglib and generates documentation at build time. This keeps the configuration reference in sync with the source of truth without manually maintained RST.

The extension falls back gracefully when the Zephyr tree is unavailable (e.g. CI) by replacing source with osource directives.

### DIFF
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -10,6 +10,7 @@ on:
     paths:
       - "include/**"
       - "docs/**"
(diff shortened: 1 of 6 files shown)


ceolin (dismissed):
Other than the comment, it looks good.

HongNguyen635 (approved):

PR #168, opened 2026-04-08
docs: improve Kconfig help text for auto-generated config reference

buckleypaul:
The Kconfig help text is the source of truth for auto-generated configuration documentation. Many entries were too terse, choice groups lacked help entirely, and it was unclear which CONFIG_ symbols are actually used in code versus which are Kconfig grouping constructs.

Expand every help entry to explain what the option does, required API calls (hubble_init, hubble_time_set, hubble_sequence_counter_get, etc.), tradeoffs, and the relationship between choice members and derived symbols like CONFIG_HUBBLE_KEY_SIZE.

Also update the kconfig_autodoc Sphinx extension to render choice-level help text so the new choice group descriptions appear in the generated docs.

Normalize indentation to single-tab throughout and fix vague prompts (e.g. "Application defined implementation" -> "Application-defined sequence counter").

### DIFF
--- a/port/zephyr/Kconfig
+++ b/port/zephyr/Kconfig
@@ -8,17 +8,28 @@ source "subsys/logging/Kconfig.template.log_config"
 choice HUBBLE_COUNTER_SOURCE
 	prompt "Counter source"
 	default HUBBLE_COUNTER_SOURCE_UNIX_TIME
+	help
+	  Selects how the SDK computes the time counter used for
+	  EID rotation and key derivation.
 
 config HUBBLE_COUNTER_SOURCE_UNIX_TIME
 	bool "Unix time counter source"
 	help
-	  Counter derived from Unix time. Requires time provisioning.
+	  Derive the time counter from Unix epoch time. The counter
+	  increments once per rotation period
+	  (CONFIG_HUBBLE_EID_ROTATION_PERIOD_SEC).
+
+	  Best for devices with access to real-time clocks or
+	  network time synchronization.
 
 config HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME
 	bool "Device uptime counter source"
 	help
-	  Counter derived from device uptime. No Unix time required.
-	  Initial counter value provided via hubble_init().
+	  Derive the time counter from device uptime. No Unix time
+	  required.
+
+	  Best for devices without a real-time clock, reliable
+	  time source, or method to provision time.
 
 endchoice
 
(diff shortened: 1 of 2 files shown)


HongNguyen635:
would it makes sense to have the kconfig of esp-idf to be matching or is it too much duplicate content?

buckleypaul:
> would it makes sense to have the kconfig of esp-idf to be matching or is it too much duplicate content?

I asked @ceolin about this a while ago and they're apparently different enough such that we can't. I agree though, it's not ideal.

port/zephyr/Kconfig
```diff
@@ -27,129 +52,198 @@ config HUBBLE_EID_ROTATION_PERIOD_SEC
...
+	  useful for unit testing where controlled time values
+	  are needed to verify time-dependent behavior.
 
 choice
 	prompt "Hubble Network crypto provider"
 	default HUBBLE_NETWORK_CRYPTO_PSA
+	help
+	  Select the cryptographic backend for AES-CTR encryption
+	  and CMAC authentication. All providers must implement
+	  the interface defined in <hubble/port/crypto.h>.
```
ceolin: This information "header" is already explained in the custom and it is not relevant for the interfaces we provided.

port/zephyr/Kconfig
```diff
@@ -27,129 +52,198 @@ config HUBBLE_EID_ROTATION_PERIOD_SEC
...
 
 config HUBBLE_SAT_NETWORK_SMALL
-	   bool "Use smaller code size for fly by calculation"
-	   help
-		Reduces code using polynomial approximation
-		for trigonometric functions
+	bool "Use smaller code size for fly-by calculation"
+	help
+	  Replace standard math library trigonometric functions
+	  (sin, cos, atan2, fmod) with polynomial approximations
```
ceolin: This level of verbosity that LLMs do is a maintenance problem, it is pretty easy to get inconsistent. We should limit to "trigonometry functions ..."

port/zephyr/Kconfig
```diff
@@ -27,129 +52,198 @@ config HUBBLE_EID_ROTATION_PERIOD_SEC
...
-	   help
-		Enable library for communication with Hubble network.
-		This feature is in pre-production and is not yet ready
-		for production deployments.
+	bool "Hubble Satellite Network Library [PRE-PRODUCTION]"
+	select EXPERIMENTAL
+	select ENTROPY_GENERATOR
+	help
+	  Enable the satellite communication module. Provides
+	  hubble_sat_packet_send() and satellite pass prediction
```
ceolin: I don't think apis should be listed. We should limit to inform that satellite functionality is provided / enabled. If we change API we have to change the symbol documentation also what is the criteria ? All APIs available with this functionality should be listed or just one ?

port/zephyr/Kconfig
```diff
@@ -8,17 +8,42 @@ source "subsys/logging/Kconfig.template.log_config"
...
+	  Best for devices with access to real-time clocks or
+	  network time synchronization.
 
 config HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME
 	bool "Device uptime counter source"
 	help
-	  Counter derived from device uptime. No Unix time required.
-	  Initial counter value provided via hubble_init().
+	  Derive the time counter from device uptime. No Unix time
+	  or network time sync required.
```
ceolin: network time ?

port/zephyr/Kconfig
```diff
@@ -8,17 +8,42 @@ source "subsys/logging/Kconfig.template.log_config"
 choice HUBBLE_COUNTER_SOURCE
 	prompt "Counter source"
 	default HUBBLE_COUNTER_SOURCE_UNIX_TIME
+	help
+	  Selects how the SDK computes the time counter used for
+	  EID rotation and key derivation. The counter determines
+	  when encryption keys and BLE addresses rotate.
```
ceolin: No, it does not necessarily determines when BLE address rotate.

port/zephyr/Kconfig
```diff
@@ -8,17 +8,42 @@ source "subsys/logging/Kconfig.template.log_config"
...
+
+	  Only the individual options below produce CONFIG_ symbols
+	  used in application and SDK source code. This choice group
+	  itself is not referenced in code.
```
ceolin: Is this paragraph really providing any meaningful / useful information ?

port/zephyr/Kconfig
```diff
@@ -8,17 +8,42 @@ source "subsys/logging/Kconfig.template.log_config"
...
 	help
-	  Counter derived from Unix time. Requires time provisioning.
+	  Derive the time counter from Unix epoch time. The counter
+	  increments once per rotation period
+	  (CONFIG_HUBBLE_EID_ROTATION_PERIOD_SEC).
+
+	  Requires providing a valid Unix timestamp in milliseconds
+	  to hubble_init(), or calling hubble_time_set() before
+	  generating advertisements. A value of 0 is treated as
+	  uninitialized and returns an error.
```
ceolin:
```
A value of 0 is treated as uninitialized and returns an error.
```

This is API documentation and not symbol documentation.

port/zephyr/Kconfig
```diff
@@ -8,17 +8,42 @@ source "subsys/logging/Kconfig.template.log_config"
...
 config HUBBLE_COUNTER_SOURCE_UNIX_TIME
 	bool "Unix time counter source"
 	help
-	  Counter derived from Unix time. Requires time provisioning.
+	  Derive the time counter from Unix epoch time. The counter
+	  increments once per rotation period
+	  (CONFIG_HUBBLE_EID_ROTATION_PERIOD_SEC).
+
+	  Requires providing a valid Unix timestamp in milliseconds
+	  to hubble_init(), or calling hubble_time_set() before
```
ceolin: APIs names are not needed here, this is part of the API documentation.

port/zephyr/Kconfig
```diff
@@ -27,129 +52,198 @@ config HUBBLE_EID_ROTATION_PERIOD_SEC
...
-		represented in PPM (parts per million).
+	int "Device time drift retry rate in PPM"
+	default 500
+	help
+	  Compensates for clock drift by adding retransmission
+	  retries proportional to time elapsed since the last
+	  time sync. Value is in parts per million (PPM).
+
+	  Additional retries = (seconds_since_sync * TDR)
+	                       / (1000000 * interval_s)
```
ceolin: Implementation detail, it should not be here.

port/zephyr/Kconfig
```diff
@@ -27,129 +52,198 @@ config HUBBLE_EID_ROTATION_PERIOD_SEC
...
-	   default y
-	   help
-		Ensure that nonce parameters are not re-used
-		to encrypt different messages.
-		Be aware that re-using the same nonce has severe
-		security implications.
+	bool "Enforce check in nonce used in cryptography"
+	default y
+	help
+	  Validate that each (time_counter, sequence_number) pair
```
ceolin: `time_counter, sequence_number` why use the variable names ??????????????
This should limit to say that it validates that the encryption nonce is unique.

port/zephyr/Kconfig
```diff
@@ -27,129 +52,198 @@ config HUBBLE_EID_ROTATION_PERIOD_SEC
...
-		This option allows applications to override the default
-		function that returns a sequence number to be used to
-		encrypt messages.
+	bool "Application-defined sequence counter"
+	help
+	  Override the default auto-incrementing sequence counter
+	  with an application-provided implementation. When
+	  enabled, the application must define
+	  hubble_sequence_counter_get() returning a uint16_t
+	  value in the range [0, 1023].
```
ceolin:
```
returning a uint16_t
	  value in the range [0, 1023].
```
API documentation ... should not be duplicated here.

ceolin (changes requested):
See individual comments please.

ceolin:
> > would it makes sense to have the kconfig of esp-idf to be matching or is it too much duplicate content?
> 
> I asked @ceolin about this a while ago and they're apparently different enough such that we can't. I agree though, it's not ideal.

The question about using the same `Kconfig` file for both targets and my answer was not because symbol dependencies are not the same or there is not guarantee that they are. e.g: all `MBEDTLS_` symbols are defined in Zephyr and so they not be the same for ESP-IDF. Also, there are things like `BUILD_WITH_TFM` and others.

ceolin (approved):

ceolin:
Thank you !

PR #169, opened 2026-04-09
api: unify key size configuration across all ports

buckleypaul:
The FreeRTOS port configured key size as a raw byte count (`CONFIG_HUBBLE_KEY_SIZE=16`), diverging from Zephyr and ESP-IDF which use human-readable choice symbols (`CONFIG_HUBBLE_NETWORK_KEY_128 / CONFIG_HUBBLE_NETWORK_KEY_256`). This inconsistency meant FreeRTOS lacked the choice symbols that mbedtls.c needs for cipher selection, and users had to know valid byte values rather than picking a key size. This was also unclear for other ports whether they should define the key size in bytes or use the `CONFIG_HUBBLE_NETWORK_KEY_128/256` values.

Centralize the byte-size derivation in crypto.h as `HUBBLE_KEY_SIZE_BYTES`, making it a computed value rather than a config knob. All ports now only define the choice symbol; the byte translation happens once.

Also rename `CONFIG_HUBBLE_KEY_SIZE` --> `HUBBLE_KEY_SIZE_BYTES` to indicate it is no longer a config value.


buckleypaul:
I'm not in love with the key size now being in hubble/port/crypto.h and needing to include that wherever we define the byte array for storing keys. However, I also don't love the implicit calculation of key bytes from the key size config values which makes porting to non-Zephyr more complicated. I'm open to suggestions.

include/hubble/port/crypto.h
```diff
@@ -10,6 +10,14 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#if defined(CONFIG_HUBBLE_NETWORK_KEY_256)
+#define HUBBLE_KEY_SIZE_BYTES 32
+#elif defined(CONFIG_HUBBLE_NETWORK_KEY_128)
+#define HUBBLE_KEY_SIZE_BYTES 16
+#else
+#error "Define CONFIG_HUBBLE_NETWORK_KEY_128 or CONFIG_HUBBLE_NETWORK_KEY_256"
+#endif
```
ceolin, line 19: Keep the name `CONFIG_HUBBLE_KEY_SIZE_BYTES` and avoid all naming changes everywhere else please. That solves the problem and avoid messing around the rest of the code. One of the reasons is because this will probably to be changed to a void pointer to allow passing an opaque type (just the key ID and not they key material) when using PSA.

ceolin (changes requested):
I think the change can be limited and avoid massive changes all around. See the comment please.

include/hubble/port/crypto.h
buckleypaul: My concern with keeping the name `CONFIG_HUBBLE_KEY_SIZE_BYTES` is that it is not a config value that we set. I'm not fully opposed to keeping that but it makes it a bit unclear if you set or don't set it. Keeping that name also doesn't solve the problem I noted?

ceolin: Which problem ?

ceolin:
@buckleypaul take a look https://github.com/HubbleNetwork/hubble-device-sdk/pull/176 please

PR #172, opened 2026-04-13
ci: add gh action to build esp-idf sample

HongNguyen635:
add github action to build esp-idf ble-beacon sample app.

### DIFF
--- a/.github/workflows/build-samples-esp-idf.yaml
+++ b/.github/workflows/build-samples-esp-idf.yaml
@@ -1,0 +1,45 @@
+name: Build ESP-IDF Ble-Beacon Sample
+
+on:
+  pull_request:
+    branches: ["**"]
+    paths:
+      - "include/**"
+      - "src/**"
+      - "samples/esp-idf/**"
+      - ".github/workflows/build-samples-esp-idf.yaml"
+
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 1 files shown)


.github/workflows/build-samples-esp-idf.yaml
```diff
@@ -0,0 +1,55 @@
...
+    name: Build ESP-IDF Ble-Beacon Sample App
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        espidf_target:
+          - esp32
+          - esp32c6
+          - esp32h2
+          - esp32s3
```
HongNguyen635, line 28: I'm not sure if we should build 1 soc or 1 soc from each family.

buckleypaul: Is there a significant difference between them? I don't see a huge negative of doing all of them.

buckleypaul (dismissed):

.github/workflows/build-samples-esp-idf.yaml
HongNguyen635: there are differences in archiecture and low level details. For BLE I don't think there's much different although I'm unsure if it might worth it to check all. Maybe could be good for catching legacy issue.

ceolin (approved):

PR #174, opened 2026-04-13
feat: add automated static analysis with cppcheck and Coccinelle CI

buckleypaul:
Add a unified `tools/analyze` script that runs cppcheck and the existing Coccinelle semantic patches against src/, include/, port/, and samples/. A new GitHub Actions workflow enforces zero new findings on pull requests.

Existing cppcheck findings are baselined in a suppressions file so CI enforces only new regressions. The coccicheck typo on line 75 is also fixed.

### DIFF
--- a/.github/workflows/static-analysis.yaml
+++ b/.github/workflows/static-analysis.yaml
@@ -12,21 +12,121 @@ on:
       - 'tools/analyze'
       - 'tools/cppcheck-suppressions.txt'
 
+permissions:
+  contents: read
+  pull-requests: write
+
 jobs:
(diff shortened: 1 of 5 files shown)


ceolin (changes requested):
-1 to prevent it from being merged in a hurry.


Too much generated machinery for something that requires deeper thoughts. Some items:

- It is not really good maintain suppression for all tools we may use in the repository.
- suppress changes should be analyzed in details

ceolin:
We do have codeql integrated in CI now. Feel free to re-open though.

PR #176, opened 2026-04-15
port: freertos: Generate CONFIG_HUBBLE_KEY_SIZE

ceolin:
Follow Zephyr a ESP-IDF behavior and change the build system to automatically generate CONFIG_HUBBLE_KEY_SIZE according with the key size selected. The symbol is added automatically to all compilation units.


buckleypaul:
This doesn't particularly solve the problem I was trying to solve in #169 - this still requires the config header to generate the key value size and this is easy to miss when moving to other implementations that don't use our ports (SiLabs, NRF-SDK, Arduino). This is an extra step for ports that doesn't feel particularly necessary when only really the crypto port needs to know the size of the key (the application should know it since it is built with the key).

src/crypto/mbedtls.c
```diff
@@ -8,16 +8,12 @@
 #include <hubble/port/sys.h>
 #include <hubble/port/crypto.h>
 
-#include "../utils/macros.h"
-
 #define _STREAM_BLOCK_LEN 16
 
-#if defined(CONFIG_HUBBLE_NETWORK_KEY_256)
```
buckleypaul, line 15: Why change this? IMO it was better before where one of those two modes needed to be defined.

include/hubble/port/crypto.h
```diff
@@ -22,6 +22,18 @@ extern "C" {
 /* Valid range [0, 1023] */
 #define HUBBLE_MAX_SEQ_COUNTER   ((1 << 10) - 1)
 
+#ifndef CONFIG_HUBBLE_KEY_SIZE
```
buckleypaul, line 25: I don't love this but it's fine. Seems like if the app wants to define it they should just use some other define name but not going to die on that hill.

include/hubble/hubble.h
```diff
@@ -67,8 +67,8 @@ extern "C" {
  *            data. The caller must ensure the buffer remains valid
  *            for the lifetime of SDK usage (do not use stack or
  *            temporary buffers). The key size must match
- *            CONFIG_HUBBLE_KEY_SIZE. If NULL, must be set with
- *            hubble_key_set before getting advertisements.
+ *            CONFIG_HUBBLE_NETWORK_KEY_256 or CONFIG_HUBBLE_NETWORK_KEY_128.
+	      If NULL, must be set with hubble_key_set before getting advertisements.
```
buckleypaul: Nit: indention

src/crypto/mbedtls.c
ceolin: it is sorted of checked in `crypto.h`, but I can revert it.

include/hubble/port/crypto.h
ceolin: It is not the app only but external ports because it is used in the function signature.

buckleypaul (approved):

PR #182, opened 2026-04-23
docs: update README

danielr-hubblenetwork:
This updates README with added details to supported platforms, links to key resources, and pointers to reference apps and sample applications to help developers get oriented quickly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,48 +1,54 @@
 # Hubble Device SDK
 
-Add global connectivity to your device using its Bluetooth chip.
+Add global connectivity to your device using its Bluetooth® chip.
 
 ## What This Does
 
-The SDK encodes your data into BLE advertisements. Hubble's gateways pick them up and deliver the data to your backend via API.
+The SDK encodes your data into Bluetooth Low Energy (BLE) advertisements that use the Hubble Service UUID. Hubble's Terrestrial Network of 90M+ gateways pick up these packets and deliver your custom payload and device location data to your backend via API.
 
-No cellular modem. No SIM card. No gateway hardware to deploy. Just Hubble firmware.
+No cellular modem. No SIM card. No gateway hardware to deploy. Just integrate the **Hubble Device SDK** with your firmware.
 
 ## Supported Platforms
 
-Works with any Bluetooth LE 5.0+ chip.
+Works with any Bluetooth LE 5.0+ chip — including hardware from Texas Instruments, Nordic, Silicon Labs, Espressif, InPlay, and any other BLE-compliant silicon.
 
-→ [Full compatibility list](https://docs.hubble.com/docs/guides/supported-devices)
+→ [Full MCU / SoC compatibility list](https://hubble.com/docs/compatibility)
 
 ## Quick Start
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,48 +1,61 @@
...
 
 Works with any Bluetooth LE 5.0+ chip.
 
-→ [Full compatibility list](https://docs.hubble.com/docs/guides/supported-devices)
+- Texas Instruments CC2340 series
+- Nordic 52 series
+- Nordic 53 series
+- Nordic 54 H and L series
+- Silicon Labs EFR32BG22
+- Espressif ESP32
```
ceolin: I don't think this list makes much sense because it is not complete. I'd say (generically) it is compatible with any Bluetooth device, but it is your call.

README.md
```diff
@@ -1,48 +1,61 @@
...
 
-2. [Register your device](https://docs.hubble.com/docs/guides/cloud-integration/register-devices) to get encryption keys
+## Sign Up
 
-3. Start transmitting
+Get started by signing up for your **Hubble Dashboard** for free: [dash.hubble.com](https://dash.hubble.com/login)
 
-→ [Reference applications](https://docs.hubble.com/docs/guides/device-integration/reference-apps) for complete working examples
+### Onboard Your First Device
+- Generate a device encryption key [via API](https://hubble.com/docs/api-specification/register-new-devices), using the [Hubble CLI](https://hubble.com/docs/guides/cli), or directly in Dash ("Add a device", then "I just need a device key").
```
ceolin: `[via API]` does not make much sense .. The `API` is kind of lose here, can we just remove it ?

README.md
```diff
@@ -1,48 +1,61 @@
...
-2. [Register your device](https://docs.hubble.com/docs/guides/cloud-integration/register-devices) to get encryption keys
+## Sign Up
 
-3. Start transmitting
+Get started by signing up for your **Hubble Dashboard** for free: [dash.hubble.com](https://dash.hubble.com/login)
 
-→ [Reference applications](https://docs.hubble.com/docs/guides/device-integration/reference-apps) for complete working examples
+### Onboard Your First Device
+- Generate a device encryption key [via API](https://hubble.com/docs/api-specification/register-new-devices), using the [Hubble CLI](https://hubble.com/docs/guides/cli), or directly in Dash ("Add a device", then "I just need a device key"). 
+- Flash your device and start beaconing on Terrestrial Network.
```
ceolin, line 35: Terrestrial Network is our terminology but it is confusing for someone that just got in here without context. If we use Terrestrial it makes sense some sort of explanation that that is our terminology for our BLE network.

ceolin (changes requested):
Minor comments that are no blocker. The -1 is just regarding the CI issue. The commit message needs to be fixed.

Thanks for the changes !

README.md
HongNguyen635: to add on to this, any BT device will work (if they support our encryption as well).
This list is more applicable to SatNet than Terrestrial.

danielr-hubblenetwork: We say " Works with any Bluetooth LE 5.0+ chip" before providing a list, which is there for scan-ability / key word search. I'll iterate on this to prevent confusion (e.g. just name a chip OEMs, not chip series, and be clear this is just a sample).

ceolin (approved):

PR #185, opened 2026-04-29
sat: ephemeris: Fix pass prediction and longitude tolerance accuracy

ceolin:
Three fixes to the satellite pass prediction logic:                                                      
                                                                                                           
- Fix loop condition in _pass_get() to advance orbit count only when both TLL crossings are in the past, preventing premature orbit advancement that caused missed or incorrect passes.
- Fix _lon_tolerance_get() spherical triangle math to use orbital radius (earth.radius + altitude) instead of altitude alone.                                            
- Replace the hardcoded satellite elevation constant with _sat_altitude_get(), which derives altitude dynamically from each satellite's mean motion and its derivative, making the longitude tolerance accurate per satellite.


HongNguyen635:
I checked out the fix. Looks like I'm still getting a few passes skipped but could be related to the recent outdated TLE causing averaging to look weird. I'll need to check for a few days to make a conclusion.
Also looks like the test is failing.

ceolin:
> I checked out the fix. Looks like I'm still getting a few passes skipped but could be related to the recent outdated TLE causing averaging to look weird. I'll need to check for a few days to make a conclusion. Also looks like the test is failing.

Can you please provide me the time used ? I have tested it yesterday and didn't see anything. The only problem that I got is that yesterday I got pass that elevation angle was so close to 30 degrees that depending on the implementation it was filter out

HongNguyen635:
After checking, this is false alarm. I apologize. I added a delta of prev_pass_ts + 80 for the pass prediction input, but the 2 consecutive passes is under 80 s.

ceolin:
> After checking, this is false alarm. I apologize. I added a delta of prev_pass_ts + 80 for the pass prediction input, but the 2 consecutive passes is under 80 s.

No need for apologies, it is better to have a false alarm than having something wrong added :)

HongNguyen635 (approved):
I ran another check with all the sats and they're all matched. So I think this is good now.

PR #187, opened 2026-04-29
sat: dtm: Add APIs to DTM

ceolin:
DTM is needed to test the operation of radio. It uses the same Bluetooth terminology and some similar APIs but it is tailored to Hubble protocol.

### DIFF
--- a/include/hubble/sat/dtm.h
+++ b/include/hubble/sat/dtm.h
@@ -1,0 +1,94 @@
+/*
+ * Copyright (c) 2026 Hubble Network, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ **/
+
+/**
+ * @file dtm.h
+ * @brief Hubble Network DTM (Direct Test Mode) APIs
+ *
+ * This module provides APIs for Direct Test Mode, which allows testing
+ * the radio physical layer by transmitting test packets on
+ * a specified channel or hopping.
+ **/
+
+#ifndef INCLUDE_HUBBLE_SAT_DTM_H
+#define INCLUDE_HUBBLE_SAT_DTM_H
+
+#include <stdint.h>
+#include <stddef.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief DTM packet type.
+ *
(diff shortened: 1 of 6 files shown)


include/hubble/sat/dtm.h
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * Selects the test packet payload size to transmit during DTM.
+ * All the payload is random bytes.
+ */
+enum hubble_sat_dtm_packet_type {
+	HUBBLE_SAT_DTM_PACKET_0,
+	HUBBLE_SAT_DTM_PACKET_4,
+	HUBBLE_SAT_DTM_PACKET_9,
+	HUBBLE_SAT_DTM_PACKET_13,
+};
```
HongNguyen635, line 37: I think this missed the -1 packet (the 1st frame of the packet only: preamble + header + 2 data).

src/hubble_sat.c
```diff
@@ -9,6 +9,7 @@
 
 #include <hubble/hubble.h>
 #include <hubble/sat.h>
+#include <hubble/sat/dtm.h>
```
HongNguyen635: should we add the `#ifdef CONFIG_HUBBLE_SAT_NETWORK_DTM_MODE` here?

include/hubble/sat/dtm.h
ceolin: it was not in my original implementation, was expecting to have it merged and then @hunterhubble  adding his changes

src/hubble_sat.c
ceolin: Not necessarily, I was thinking more about adding stubs for the API when it is not defined than ifdef the code.

But the way it is won’t cause any problem, but it is probably better to ifdef it.

include/hubble/sat/dtm.h
HongNguyen635: Okay. We could merge this in 1 go with the co-author added, but either way works for me as long as we'll include it later.

ceolin: of course

HongNguyen635 (approved):
lgtm

PR #188, opened 2026-04-30
ci: add gh action to build TI sample

HongNguyen635:
add github action to build TI ble-beacon sample app

### DIFF
--- a/.github/workflows/build-sampes-ti.yaml
+++ b/.github/workflows/build-sampes-ti.yaml
@@ -1,0 +1,36 @@
+name: Build TI Samples
+
+on:
+  pull_request:
+    branches: ["**"]
+    paths:
+      - "include/**"
+      - "src/**"
+      - "samples/freertos/ti/**"
+      - ".github/workflows/build-sampes-ti.yaml"
+
+  workflow_dispatch: {}
+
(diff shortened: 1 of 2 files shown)


ceolin (changes requested):
I am ok with these changes, but the commit message in https://github.com/HubbleNetwork/hubble-device-sdk/pull/188/changes/82dc99e21230864b64a19ac3d0ca2429e464929a is not right. The problem was not introduced there, actually that change fixed the build issue for the previous TI SDK version.

Unfortunately this dependency seems to be needed by default with the SDK recently released. This information should be in the commit message so we can understand these changes.

HongNguyen635:
> I am ok with these changes, but the commit message in [82dc99e](https://github.com/HubbleNetwork/hubble-device-sdk/commit/82dc99e21230864b64a19ac3d0ca2429e464929a) is not right. The problem was not introduced there, actually that change fixed the build issue for the previous TI SDK version.
> 
> Unfortunately this dependency seems to be needed by default with the SDK recently released. This information should be in the commit message so we can understand these changes.

Good call, thank you. I added the explaination and the version in the commit message for reference later.

ceolin (approved):

PR #189, opened 2026-05-01
ci: esp-idf: build for all esp-idf samples with footprint report

HongNguyen635:
Add footprint report to esp-idf port.

The plain-text will look like this:

```
hubblenetwork-sdk Footprint Summary
=====================================================================================
Archive:         libhubblenetwork-sdk.a
Revision:        c67feb9
ESP-IDF Version: ESP-IDF v6.1-dev-4427-gc00874869b
Run Date:        2026-05-01 23:24:01 UTC
=====================================================================================
Sample                        Target                   ROM            RAM            
-------------------------------------------------------------------------------------
ble-beacon                    esp32                    
                              Hubblenetwork-SDK:       2.21 KB        43.00 B        
                                  port/esp-idf:        7.00 B         0.00 B         
                                  port/freertos:       174.00 B       12.00 B        
ble-beacon                    esp32c6                  
                              Hubblenetwork-SDK:       2.63 KB        43.00 B        
                                  port/esp-idf:        20.00 B        0.00 B         
                                  port/freertos:       212.00 B       12.00 B        
ble-beacon                    esp32h2                  
                              Hubblenetwork-SDK:       2.63 KB        43.00 B        
                                  port/esp-idf:        20.00 B        0.00 B         
                                  port/freertos:       212.00 B       12.00 B        
ble-beacon                    esp32s3                  
                              Hubblenetwork-SDK:       2.21 KB        43.00 B        
                                  port/esp-idf:        7.00 B         0.00 B         
                                  port/freertos:       182.00 B       12.00 B        
-------------------------------------------------------------------------------------

Summary: 4 build(s) processed
=====================================================================================
```

And the Action Summary:

Screenshot 2026-05-01 at 4 26 57 PM

### DIFF
--- a/.github/workflows/build-samples-esp-idf.yaml
+++ b/.github/workflows/build-samples-esp-idf.yaml
@@ -1,4 +1,4 @@
-name: Build ESP-IDF Ble-Beacon Sample
+name: Build ESP-IDF Samples
 
 on:
   pull_request:
@@ -16,7 +16,7 @@ permissions:
 
 jobs:
   build:
-    name: Build ESP-IDF Ble-Beacon Sample App
+    name: Build ESP-IDF Sample Apps
     runs-on: ubuntu-24.04
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


tools/ci/summarize_footprint_esp_idf.py
```diff
@@ -0,0 +1,330 @@
...
+
+# Port directories that contributes to esp-idf
+PORT_DIRS = (
+    "port/esp-idf",
+    "port/freertos",
+)
+SDK_COMPONENT_ARCHIVE = "libhubblenetwork-sdk.a"
+
+SIZE_COMPONENTS_FROM_FILENAME = re.compile(r"size-components-(.+)-([^-]+)\.json$")
+SIZE_FILES_FROM_FILENAME = re.compile(r"size-files-(.+)-([^-]+)\.json$")
```
ceolin: Is this ever used ?

tools/ci/summarize_footprint_esp_idf.py
```diff
@@ -0,0 +1,330 @@
...
+        "Revision,ESP-IDF Version"
+    )
+
+    def s(v):
+        return str(v) if v is not None else "N/A"
+
+    for sample, target, rom, ram, ports in entries:
+        print(
+            f"{sample},{target},total,"
+            f"{s(rom)},{s(ram)},"
```
ceolin, line 230: Why do you have it in here and not in the other formats ? It seems redundant because you are already printing it using `format_size`. Without this you can remove that `def s(v)` function.

.github/workflows/build-samples-esp-idf.yaml
```diff
@@ -32,14 +32,91 @@ jobs:
...
+               esp-idf-footprint-out/idf-version.txt
+          fi
+
+      - name: Upload footprint artifact
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: footprint-${{ matrix.espidf_target }}
+          path: esp-idf-footprint-out/
+          if-no-files-found: ignore
```
ceolin: Shouldn't we generate a warning instead of just ignoring ? `s/ignore/warn/`

ceolin (changes requested):
Minor comments, it is looking great :)

tools/ci/summarize_footprint_esp_idf.py
HongNguyen635: oops I added to do both search but then later removed it because I only need 1 to figure out the sample and target name.

tools/ci/summarize_footprint_esp_idf.py
HongNguyen635: I was thinking about since it's csv, maybe people can do sorting on the column. But the formatted one includes characters so might not be easy.

.github/workflows/build-samples-esp-idf.yaml
HongNguyen635: I was thinking since this step will always run, even if the build fails, then we can just ignore if there is no file to upload. But I can add the warning.

ceolin (approved):

PR #193, opened 2026-05-05
zephyr: boards: add satellite support for Silabs efr32xg24

HongNguyen635:
- Add satellite support for efr32xg24.
- Add 2 more silabs dev boards support in `sat-continuous` sample.


port/zephyr/boards/silabs/efr32/radio.c
```diff
@@ -0,0 +1,245 @@
...
+					 SL_RAIL_TX_OPTIONS_DEFAULT);
+	if (status != SL_RAIL_STATUS_NO_ERROR) {
+		return status;
+	}
+
+	anchor = sl_rail_get_time(_rail_handle);
+	status = sl_rail_set_timer(_rail_handle, anchor + duration_us,
+				   SL_RAIL_TIME_ABSOLUTE, &_timer_cb);
+	if (status != SL_RAIL_STATUS_NO_ERROR) {
+		return status;
```
ceolin: If this happens the tx stream is left running, it should be stopped before returning, or even better is to set the timer before starting the stream.

port/zephyr/boards/silabs/efr32/radio.c
```diff
@@ -0,0 +1,245 @@
...
+#include <hubble/port/sat_radio.h>
+
+#include <sl_rail.h>
+#include <sl_rail_util_pa_conversions.h>
+
+#include "rail_config.h"
+
+/* EFR32 step size = 72 - 74 Hz */
+#define EFR32_STEP_SCALE(step) (step * 4)
+#define MAX_POWER_DEBM         200 /* DBm = 10 * dBm */
```
ceolin: typo ? `DBM` or `DDBM` ?

ceolin (changes requested):
No major comments, the pr is looking great.

My main concern is about propagating `sl_rail_status_t` in case of error, our patter is to return `0` in case of success and a negative value (preference to errno) otherwise. If `sl_rail_status_t` is positive, any place doing something like

`if (ret < 0)` will not behave properly. The best thing to do is something similar to https://github.com/HubbleNetwork/hubble-device-sdk/blob/b1dc504d885bfabef6d48c168d91df1bf3f7b418/src/crypto/psa.c#L10

port/zephyr/boards/silabs/efr32/radio.c
HongNguyen635: It was meant for "Deci"-dbm but since the macro is capitalized, we can't differentiate between if DBM means dBm or DBm. Looks weird though, I'll change to DDBM.

HongNguyen635:
Thank you @ceolin, I added the function to convert from [RAIL errors](https://github.com/zephyrproject-rtos/hal_silabs/blob/0b359e73c0080ffc03116bf85d8bb1b385472142/simplicity_sdk/platform/radio/rail_lib/common/sl_rail_types.h#L145) to errno. Please take a look again.

ceolin (approved):

ceolin:
> Thank you @ceolin, I added the function to convert from [RAIL errors](https://github.com/zephyrproject-rtos/hal_silabs/blob/0b359e73c0080ffc03116bf85d8bb1b385472142/simplicity_sdk/platform/radio/rail_lib/common/sl_rail_types.h#L145) to errno. Please take a look again.

Thank you !

PR #194, opened 2026-05-07
zephyr: boards: nrf54: clear RADIO SUBSCRIBE_RXEN on SoC enable

ceolin:
In dual-stack configurations, the Bluetooth controller configures a DPPI subscription on RADIO.RXEN. If a BLE connection has been established, that subscription persists when the satellite stack takes over the radio, causing spurious RX enables once the radio is reconfigured for satellite operation.

Clear SUBSCRIBE_RXEN before applying the satellite radio configuration so the radio starts from a clean state regardless of prior BLE activity.


HongNguyen635 (approved):
I checked out the fix and it works now

hunterhubble (approved):

PR #195, opened 2026-05-07
sat: add debug and warning logs for transmission and ephemeris

ceolin:
Add useful debug / warning messages around satellite common code.

### DIFF
--- a/src/hubble_sat_ephemeris.c
+++ b/src/hubble_sat_ephemeris.c
@@ -1,4 +1,5 @@
 #include <hubble/sat/ephemeris.h>
+#include <hubble/port/sys.h>
 
 #include <math.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


hunterhubble (approved):

src/hubble_sat_ephemeris.c
HongNguyen635, line 642: would it be helpful if we can log which sat it succesfully computed for?

HongNguyen635 (approved):
A minor comment but looks good to me.

src/hubble_sat_ephemeris.c
ceolin: There is a following log for success, and a valid pass here is not necessarily the returned value so it can be misleading. The sat id info should be only for the returned pass

PR #197, opened 2026-05-08
sat: use hubble_internal_time_drift_get() for drift compensation

ceolin:
Add an internal API to compute the device drift based on TDR to shift computed pass start back by half the drift window in hubble_next_pass_get().

### DIFF
--- a/src/hubble_sat_ephemeris.c
+++ b/src/hubble_sat_ephemeris.c
@@ -499,14 +499,14 @@ static int _next_pass_get(const struct hubble_sat_orbital_params *orbit,
 	}
 
 	/* Iterate until a valid pass is found */
-	while (pass->t == 0 &&
+	while (pass->culmination == 0 &&
 	       (HUBBLE_TWO_PI_DEGREES -
 			_zero_to_360(pos->lon - lon_tol - crossings[index].lon) <
 		HUBBLE_PI_DEGREES)) {
 		if ((fabs(_minus_180_to_180(crossings[index].lon - pos->lon)) <=
 		     lon_tol) &&
 		    (crossings[index].t > t)) {
-			pass->t = crossings[index].t;
+			pass->culmination = crossings[index].t;
 			pass->lon = crossings[index].lon;
 			pass->ascending =
 				(ascending) ? pos->lat > 0 : pos->lat <= 0;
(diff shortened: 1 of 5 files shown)


src/hubble_sat_ephemeris.c
```diff
@@ -653,6 +655,15 @@ int hubble_next_pass_get(uint64_t t, const struct hubble_sat_device_pos *pos,
...
 	}
 
+	/* Compensate for clock drift: the device may be ahead or behind by up
+	 * to hubble_internal_time_drift_get() ms, so start half a drift-window
+	 * early to keep the pass centered on the actual reception window.
+	 *
+	 * hubble_internal_time_drift_get() returns ms so we need make it
+	 * seconds and divide by 2.
+	 */
+	pass->t -= hubble_internal_time_drift_get() / 2000U;
```
HongNguyen635: Just 1 point: this will only work if the drift if ahead right? If the drift is behind, and we move the pass window further back, we'll miss the pass?

HongNguyen635 (dismissed):
Just 1 minor comment, otherwise this looks good to me.

src/hubble_sat_ephemeris.c
ceolin: I am glad you noticed it, I was debating about it and decided to consider a valid pass if we have the whole window, note that this will still return the pass and it will be application decision whether or not using it.

src/hubble_sat_ephemeris.c
```diff
@@ -778,23 +799,33 @@ int hubble_next_pass_region_get(uint64_t t,
...
 				   "pass found across %zu satellites",
 				   _satellites_count);
 		return -ENOENT;
 	}
 
-	pass->t -= pass->duration / 2;
+	transmission_period_s =
+		hubble_internal_sat_transmission_period_get() / 1000U;
+
+	pass->duration += transmission_period_s;
```
HongNguyen635, line 817: I feel like I'm missing sometihng. This somehow doesn't feel right to me. For region, we already have the duration that sat passes over a region right? `hubble_internal_sat_transmission_period_get` will give you the number of retries. But this total retries is not right. Aren't we supposed to somehow do this:
`total retries = duration / retries_period`?

For single point prediction, it makes sense to call `hubble_internal_sat_transmission_period_get`, but for a region, we know that the retries period (even with no clock drift) is 160s, and the duration for a region could be longer than that.

Since we don't control the transmission period for region, I feel like the math for single point prediction doesn't apply here.

ceolin: We don't know the number of retries but we need to accommodate the situation were the device is located in the boundary of the region, if the device is in the boundary we can't simply wake up in the culmination time. This applies the same semantics, splitting the period of a single point transmission between the beginning and end the of the region. Remember that we calculate the duration based on the culmination time of when the sat enters and exits the region.

HongNguyen635: right you mentioned about the culmination time and I keep forgetting that the time sat enter/exit is at culmination. This makes sense then.

HongNguyen635 (approved):

PR #198, opened 2026-05-18
sat: ephemeris: refine pass culmination

ceolin:
Add _next_pass_culmination_get() which applies a single Newton-Raphson step over the linearized satellite-observer relative motion (position, velocity, and centripetal acceleration) to correct the coarse culmination time estimate. Update test expected values to match the improved predictions.


HongNguyen635 (commented):
Looks like there is a bug.
At the current time when I tried this out, if I use the `predicted culmination time + t` as the input for next pass, a few scenarios:
- For t <= 20 s , it returns the same pass.
- For 20 < t < 25 s, repeated results for future passes.
- For t >= 25, results is normal and great.

ceolin:
> Looks like there is a bug. At the current time when I tried this out, if I use the `predicted culmination time + t` as the input for next pass, a few scenarios:
> 
> * For t <= 20 s , it returns the same pass.
> * For 20 < t < 25 s, repeated results for future passes.
> * For t >= 25, results is normal and great.

culmination is computed after the computation of the pass prediction time, depending on the delta between pass prediction `pT` and culmination `cT`, the logic in the application `predicted culmination time + t` can return a value lower that is lower than `pT` and you ended up getting the same report. Applications should do `pass->start + pass->duration` or `pass->culmination + pass->duration`.

HongNguyen635 (approved):
@ceolin that worked. We should put a line in the doc comment, something like:
`if the input time is in the middle of the pass, that pass will be returned, and start time is in the past`
or something like that.

PR #199, opened 2026-05-20
samples: ti/ble-beacon: Fix lib rename after update to 9.20.00.81

HongNguyen635:
`common_iCall_icall_cc23x0` has been renamed to `common_iCall_icall` in TI v9.20.00.81


ceolin (approved):
LGTM !

PR #201, opened 2026-05-22
ci: build all TI samples instead of just the BLE beacon

HongNguyen635:
Introduce a different way to build for TI:
- Have a `sample.yaml` in each sample dir with different configs we want to build.
- Then the workflow will use the script to recursively search in the samples dir and build these.
- We don't need to manually adding entry for every sample, every config anymore.

### DIFF
--- a/samples/freertos/ti/ble-beacon/makefile
+++ b/samples/freertos/ti/ble-beacon/makefile
@@ -94,6 +94,12 @@ LFLAGS += -Wl,--diag_wrap=off \
 	  $(BUILD_DIR)/cc2340_freertos.cmd \
 	  -llibc.a
 
+# check if key and time is generated
+generated_files = src/key.c src/time.c
+ifeq ($(words $(wildcard $(generated_files))),2)
+	CFLAGS += -DHUBBLE_KEY_TIME_SET
+endif
+
 all: $(BUILD_DIR) postbuild
 
 $(BUILD_DIR):
--- a/samples/freertos/ti/ble-beacon/sample.yaml
+++ b/samples/freertos/ti/ble-beacon/sample.yaml
@@ -1,0 +1,8 @@
+sample:
+  name: ble-beacon
+
+builds:
+  - id: default
+    makefile: makefile
+  - id: SysConfig-based
+    makefile: makefile-syscfg
--- a/tools/ci/ti_build.py
+++ b/tools/ci/ti_build.py
@@ -1,0 +1,360 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2026 Hubble Network, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Recursive search for sample.yaml files under a directory, and build each sample configuration.
+
+The sample.yaml format is as follows:
+
+.. code-block:: yaml
+
+    sample:
+      name: <optional, sample name, defaults to parent directory name>
+
+    builds:
+      - id: <build identifier (string), required>
+        makefile: <path to makefile, required>
+        extra_args: <optional, list of extra arguments passed to make>
+"""
+
+from __future__ import annotations
+
+import argparse
+import os
+import shlex
+import subprocess
+import sys
+import time
+from dataclasses import dataclass, field
+from pathlib import Path
+
+import yaml
+
+
+@dataclass
+class Build:
+    """
+    Single build configuration for a sample.
+    """
+
+    sample_name: str
+    sample_dir: Path
+    build_id: str
+    makefile: str
+    extra_args: list[str] = field(default_factory=list)
+
+    @property
+    def label(self) -> str:
+        return f"{self.sample_name} / {self.build_id}"
+
+
+@dataclass
+class Sample:
+    """
+    Container for 1 sample, which may have multiple build configurations.
+    """
+
+    name: str
+    sample_dir: Path
+    builds: list[Build] = field(default_factory=list)
+
+
+def parse_sample_yaml(yaml_path: Path) -> Sample:
+    """
+    Parse a sample.yaml file and return a Sample object.
+
+    Args:
+        yaml_path: Path to the `sample.yaml` file.
+
+    Returns:
+        Sample object with the parsed configuration.
+    """
+    with yaml_path.open() as f:
+        data = yaml.safe_load(f) or {}
+
+    # extract the name or default to parent dir name
+    sample_meta = data.get("sample") or {}
+    name = str(sample_meta.get("name") or yaml_path.parent.name)
+    sample_dir = yaml_path.parent
+
+    # get build options
+    builds_raw = data.get("builds")
+    if not builds_raw:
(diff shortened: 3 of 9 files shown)


HongNguyen635:
Example of build summary. The build will stop if a sample fails to build midway.

Screenshot 2026-05-26 at 7 26 39 AM

tools/ci/ti_build.py
```diff
@@ -0,0 +1,404 @@
...
+        if not isinstance(entry, dict):
+            raise ValueError(f"{yaml_path}: each build entry must be a mapping")
+        if "id" not in entry:
+            raise ValueError(f"{yaml_path}: build entry missing 'id'")
+        if "makefile" not in entry:
+            raise ValueError(
+                f"{yaml_path}: build entry '{entry['id']}' missing 'makefile'"
+            )
+
+        build_id = str(entry["id"])
```
ceolin: Isn't it already a string ?

tools/ci/ti_build.py
```diff
@@ -0,0 +1,404 @@
...
+            if not success:
+                print(output, end="", flush=True)
+                print(f"::error::Build failed: {build.label}", flush=True)
+                stop = True
+                break
+        if stop:
+            break
+
+    failed_builds = [build for build, success, _ in results if not success]
+    passed_count = sum(1 for _, success, _ in results if success)
```
ceolin: not very efficient this .. you could assign both variable doing one single pass here.

tools/ci/ti_build.py
```diff
@@ -0,0 +1,404 @@
...
+
+    if failed_builds:
+        print(flush=True)
+        print("Failed builds:", flush=True)
+        for build in failed_builds:
+            print(f"  - {build.label}", flush=True)
+
+    write_build_summary(results, total, prebuild_failed)
+
+    return 1 if failed_builds or prebuild_failed else 0
```
ceolin: Why are you forcing to flush all prints ? What is the need for that ?

If you really need it everywhere why don't simply execute this script with `-u` option ?

ceolin (changes requested):
Minor things, it is looking super good

tools/ci/ti_build.py
HongNguyen635: I read online and it said that flushing will let us see the progress immediately instead of buffered it and dumps data after sometime. But you're right, that might be just overkill for this one. I was debating whether to include it or not and I think I'll just remove it.

tools/ci/ti_build.py
HongNguyen635: ah good catch. At first I was thinking of assigning numbers to each build and then changed my mind to give it better name.

HongNguyen635:
I addressed the comments, please take a look again.

samples/freertos/ti/ble-beacon/sample.yaml
```diff
@@ -0,0 +1,12 @@
+sample:
+  name: ble-beacon
+
+prebuild:
+  - run: echo -n "AAAAAAAAAAAAAAAAAAAAAA==" > master.key
+  - run: python3 tools/embed_key_time.py -b master.key -o samples/freertos/ti/ble-beacon/src/
```
ceolin: @HongNguyen635 This can't be allowed. The action runs in the pull request context, this means that the user controls it and we are running whatever is set in here.

The easiest way is to set the fake key and dummy time directly in the code adding some comments that it must be overwrite with proper values.

ceolin (changes requested):
There is a security implication between the script and the gh action.

samples/freertos/ti/ble-beacon/sample.yaml
HongNguyen635: that's a good point.

HongNguyen635:
@ceolin that's a valid concern, and it's true for the `builds` step as well. Realistically speaking, attacker could put anything in the `make` file and we execute this make file on pull requests.

Since the scope is limited to read, they can't read repo secrets, but they can burn CI time for crypto mining for example. How do we want to handle this? Do you want to limit this workflow to trigger manually with write access reviewers?

samples/freertos/ti/ble-beacon/makefile
```diff
@@ -94,6 +94,12 @@ LFLAGS += -Wl,--diag_wrap=off \
 	  $(BUILD_DIR)/cc2340_freertos.cmd \
 	  -llibc.a
 
+# check if key and time is generated
+generated_files = src/key.c src/time.c
+ifeq ($(words $(wildcard $(generated_files))),2)
+	CFLAGS += -DHUBBLE_KEY_TIME_SET
```
HongNguyen635, line 100: I'm not sure if there is an alternative way. I was thinking of using `__has_include` but this is compiler dependent...

tools/ci/ti_build.py
```diff
@@ -0,0 +1,370 @@
...
+
+        extra_args = entry.get("extra_args") or []
+        if not isinstance(extra_args, list):
+            raise ValueError(
+                f"{yaml_path}: 'extra_args' for '{build_id}' must be a list"
+            )
+
+        # Validate extra_args does not contain `-`, which can
+        # redirect / inject arbitrary options.
+        extra_args = [str(a) for a in extra_args]
```
ceolin: I think you should drop it, it just gives a false security sensation, one can easily pass it setting to extra_args something like `"SHELL=/bin/sh -c \"…\""` or using `MAKEFLAGS` ...

samples/freertos/ti/ble-beacon/makefile
ceolin: It is fine using compiler attributes for sample tight to specific SDKs/toolchains, but I prefer the way you did it :)

tools/ci/ti_build.py
```diff
@@ -0,0 +1,370 @@
...
+        pct = (passed_count / total * 100) if total else 0.0
+        print(
+            f"{passed_count} of {total} builds passed ({pct:.1f}%), {len(failed_builds)} failed."
+        )
+
+    if failed_builds:
+        print()
+        print("Failed builds:")
+        for build in failed_builds:
+            print(f"  - {build.label}")
```
ceolin, line 352: Isn't all this print logic duplicated from `write_build_summary` ? can't print there along with the file writing ?

ceolin (changes requested):
Minor changes and we can merge it.

There is a still a possibility of exploiting CI resources but no information leakage because of the action permissions. I think it is fine - at least for now :)

tools/ci/ti_build.py
HongNguyen635: So my intention is the `write_build_summary` is specific for GH summary, while with the print statement, I can run this script locally and still see the output. There's a little bit of duplicate, I originally check in `write_build_summary` for if the env variable exists, else just print to stdout, but it's a bit hard to read because of GH format, so I just print to stdout.

ceolin (approved):

PR #204, opened 2026-05-26
sat: ephemeris: Small fixes and improvements

ceolin:
- Optimizations to avoid `libm` when `CONFIG_HUBBLE_SAT_NETWORK_SMALL` is set
- Fix possible underflow
- Use sidereal instead of solar day rotation

### DIFF
--- a/src/hubble_sat_ephemeris.c
+++ b/src/hubble_sat_ephemeris.c
@@ -126,7 +126,7 @@ static void _range_reduce(double x, int *q, double *r)
 {
 	/* Reduce x/π to integer multiple of 0.5 */
 	double n = nearbyint(
-		x * (0.5 * HUBBLE_INV_PI)); /* round to nearest multiple of π/2 */
+		x * (2.0 * HUBBLE_INV_PI)); /* round to nearest multiple of π/2 */
 
 	*q = (int)n;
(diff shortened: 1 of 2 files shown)


src/hubble_sat_ephemeris.c
```diff
@@ -288,6 +288,62 @@ static inline double _asin_small(double x)
 	return _atan_small(x / denom);
 }
 
+/* Extracts log2 of a positive normal double via IEEE 754 exponent + atanh series */
+static double _log2_small(double x)
```
buckleypaul, line 292: This is a pretty complicated function with no testing. We should either add testing for it and/or document where it came from (that presumably is tested).

ceolin: It is fair ... I have tested comparing pass prediction with and without `HUBBLE_SAT_NETWORK_SMALL`. I don't think it needs dedicated unit tests for now, but I will ensure that the pass prediction test exercises this.

HongNguyen635 (dismissed):
Just to add to Paul's comment above, should we add the `CONFIG_ HUBBLE_SAT_NETWORK_SMALL` to the ephemeris test case?

HongNguyen635 (approved):

PR #206, opened 2026-05-26
sat: dtm: Add -1 payload mode

hunterhubble:
Add -1 payload mode to only send the first 16 symbols (8 preamble, 6+2 header).

### DIFF
--- a/src/hubble_sat.c
+++ b/src/hubble_sat.c
@@ -159,6 +159,10 @@ int hubble_sat_dtm_packet_send(enum hubble_sat_dtm_packet_type type,
 	uint8_t len;
 	int ret;
 
+	if (type == HUBBLE_SAT_DTM_PACKET_SINGLE_FRAME) {
+		packet.length = HUBBLE_SAT_DTM_PACKET_ONE_FRAME_ONLY_LEN;
+		goto end;
+	}
 	switch (type) {
 	case HUBBLE_SAT_DTM_PACKET_0:
 		len = 0;
@@ -184,6 +188,7 @@ int hubble_sat_dtm_packet_send(enum hubble_sat_dtm_packet_type type,
 		return ret;
 	}
 
+end:
 	ret = hubble_sat_dtm_port_packet_send(&packet, channel);
 	if (ret < 0) {
 		HUBBLE_LOG_WARNING(
--- a/src/hubble_sat_packet.c
+++ b/src/hubble_sat_packet.c
@@ -12,6 +12,10 @@
 #include <hubble/port/sat_radio.h>
 #include <hubble/port/sys.h>
 
+#ifdef CONFIG_HUBBLE_SAT_NETWORK_DTM_MODE
+#include <hubble/sat/dtm.h>
+#endif
+
 #include "reed_solomon_encoder.h"
 #include "hubble_priv.h"
 #include "utils/bitarray.h"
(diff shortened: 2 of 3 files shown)


src/hubble_sat.c
```diff
@@ -159,6 +159,11 @@ int hubble_sat_dtm_packet_send(enum hubble_sat_dtm_packet_type type,
 	uint8_t len;
 	int ret;
 
+	if (type == HUBBLE_SAT_DTM_PACKET_SINGLE_FRAME) {
+		packet.length = HUBBLE_SAT_DTM_PACKET_ONE_FRAME_ONLY_LEN;
+		len = 0;
```
ceolin: there is no need for this assignment since it is not used when this branch executes.

src/hubble_sat_packet.c
```diff
@@ -393,6 +394,15 @@ int hubble_sat_packet_frames_get(const struct hubble_sat_packet *packet,
 	}
 	frame_pos += HUBBLE_PHY_ECC_SYMBOLS_SIZE;
 
+#ifdef CONFIG_HUBBLE_SAT_NETWORK_DTM_MODE
+	/* Payload -1 mode will cause an early return */
+	if (packet->length == HUBBLE_SAT_DTM_PACKET_ONE_FRAME_ONLY_LEN) {
+		frames->total_number_of_symbols = HUBBLE_PACKET_FRAME_PAYLOAD_MAX_SIZE;
```
ceolin: that does not look right since the port will use it to send the data. You probably needs to set `frames->total_number_of_symbols = frame_pos;` don't you ?

src/hubble_sat_packet.c
```diff
@@ -11,39 +11,40 @@
...
 /* Number of bits to represent sequency number*/
-#define HUBBLE_SEQUENCE_NUMBER_SIZE          10U
+#define HUBBLE_SEQUENCE_NUMBER_SIZE              10U
 
-#define HUBBLE_PAYLOAD_MAX_SIZE              13U
+#define HUBBLE_PAYLOAD_MAX_SIZE                  13U
 
 /* Two bits used -> 2**2 */
-#define HUBBLE_HOPPING_SEQUENCE_INFO_NUM     4U
+#define HUBBLE_HOPPING_SEQUENCE_INFO_NUM         4U
```
ceolin: you haven't added or changed any of those symbols so don't change the identation.

ceolin (changes requested):
A few comments / changes and it will be good to get in.

src/hubble_sat_packet.c
hunterhubble: After testing, it seems that using frame_pos only sent 14 symbols, when we want the whole 16. We don't just want the preamble and header, we want the entire frame to show what one channel hop is, so there's two more 0 value symbols at the end.

src/hubble_sat.c
hunterhubble: Noted, thanks

src/hubble_sat_packet.c
hunterhubble: I see. I ran clang-format when I was still in hubblenetwork-sdk, but didn't do it after applying the diffs from that commit to hubble-device-sdk. I fixed that just now, thank you!

src/hubble_sat_packet.c
```diff
@@ -11,6 +11,7 @@
 #include <hubble/sat/packet.h>
 #include <hubble/port/sat_radio.h>
 #include <hubble/port/sys.h>
+#include <hubble/sat/dtm.h>
```
HongNguyen635, line 16: Since we're doing this below, we can probably add:

```
#ifdef CONFIG_HUBBLE_SAT_NETWORK_DTM_MODE

#endif
```

HongNguyen635 (commented):
minor comment, otherwise LGTM

src/hubble_sat_packet.c
hunterhubble: Done, should that also be done for hubble_sat.c as well though?

HongNguyen635: this [line](https://github.com/HubbleNetwork/hubble-device-sdk/blob/7881429daaf392a5ab9d788d22bf9a82fcef8c1b/src/hubble_sat.c#L15)?

HongNguyen635 (approved):
Looking great, thank you!

src/hubble_sat_packet.c
ceolin: That makes sense to me, thanks for clarifying it.

ceolin (approved):

PR #208, opened 2026-05-28
fix: ble-beacon/freertos: Avoid heap exhaustion

### LINKED ISSUE #203: TI sample - fails after 20 packet refreshes
### Hubble Network  SDK Commit Hash

56f86246ea393f49f26dd79853f00bb97abfdda7

### platform-port

Free RTOS Ti 

### Hardware Target

CC2745

### Steps to reproduce

After not knowing why the board was not advertising after a while - in my case was ~20hr, turns out the adv_set was being recreated every 1hr and not freeing the old one, so then just went silent after running out of 20x adv sets - with some Ai checking of the code, here is the summary and fix.  Note that was able to accelerate the packet refresh time to 5sec to quickly prove the bug was there, then fix it.  Here is the Claude breakdown: 

— AI-assisted writeup below —

Affected: samples/freertos/ti/ble-beacon/src/hubble_ble_adv.c → hubble_ble_adv_update(). Repros on a CC2745 dev kit, SimpleLink Low Power F3 SDK 9.14. Confirmed on v2.0.0 (56f86246) and still on current main (531b6f9).

How to reproduce
Build and flash the unmodified ble-beacon sample to a CC2745 dev kit.
Open a BLE scanner (e.g. nRF Connect) and filter on the Hubble service UUID 0xFCA6. Confirm the beacon is advertising.
Leave it running. With the sample default HUBBLE_ADV_PACKET_PERIOD = 180000000UL (180 s), the beacon stops advertising after ~20 refreshes ≈ 1 hour and never returns (only a reboot restarts it — then it dies again on the same clock).
To repro in ~2 minutes instead of an hour: set HUBBLE_ADV_PACKET_PERIOD = 5000000UL (5 s) and rebuild. It now goes silent at ~20 × 5 s.
On my CC2745 dev kit it stopped at exactly 1:40 (100 s). The clean "exactly 20" points to the advertising-set handle ceiling (AE_MAX_NUM_ADV_SETS = 20), not heap exhaustion.
On serial/RTT you'll see the sample's own "Failed to initialize advertising set" at the moment it dies — that's GapAdv_create() failing.
Why: each refresh, hubble_ble_adv_update() calls BLEAppUtil_initAdvSet() → GapAdv_create(), allocating a new advertising set with no matching GapAdv_destroy(). After ~20 it runs out of handles, and the failing create happens after advStop already disabled the radio — with no retry — so advertising never resumes.

The fix
Update the existing set's data in place instead of tearing it down and rebuilding it (the set is already created once at startup in hubble_ble_adv_start()).

In hubble_ble_adv_update(), replace:

```
	if (BLEAppUtil_advStop(bleAdvHandle) != SUCCESS) {
		Log_printf(LogModule_Beacon, Log_ERROR, "Failed to stop advertising");
		return;
	}
	if (BLEAppUtil_initAdvSet(&bleAdvHandle, &hubbleInitAdvSet) != SUCCESS) {
		Log_printf(LogModule_Beacon, Log_ERROR, "Failed to initialize advertising set");
		return;
	}
	(void)BLEAppUtil_advStart(bleAdvHandle, &hubbleStartAdvSet);
```
with:

```
	/* Update the existing advertising set in place; do not re-create it.
	 * DONT_FREE: advData is a static buffer. The host auto-re-enables
	 * advertising after the load, so no advStart() is needed. */
	if (GapAdv_prepareLoadByHandle(bleAdvHandle, GAP_ADV_FREE_OPTION_DONT_FREE) != SUCCESS) {
		Log_printf(LogModule_Beacon, Log_ERROR, "Failed to prepare adv data load");
		return;
	}
	if (GapAdv_loadByHandle(bleAdvHandle, GAP_ADV_DATA_TYPE_ADV, sizeof(advData), advData) != SUCCESS) {
		Log_printf(LogModule_Beacon, Log_ERROR, "Failed to load adv data");
		return;
	}
```
No per-refresh GapAdv_create(), so nothing accumulates. Verified on my CC2745 dev kit: with the 5 s refresh kept, it now runs well past the old 100 s failure point. 

### Logs, anything relevant to understand the problem

```shell

```

### Impact

Major

ceolin:
Replace stop/reinit/start sequence with GapAdv_prepareLoadByHandle and GapAdv_loadByHandle to update advertisement data in-place, avoiding advertising gaps during data updates and heap exhaustion.

Fixes #203

### DIFF
--- a/samples/freertos/ti/ble-beacon/src/hubble_ble_adv.c
+++ b/samples/freertos/ti/ble-beacon/src/hubble_ble_adv.c
@@ -110,19 +110,19 @@ static void hubble_ble_adv_update(void *arg)
 
 	advData[4] = len + 1; /* output len + BLE service data type */
 
-	if (BLEAppUtil_advStop(bleAdvHandle) != SUCCESS) {
+	if (GapAdv_prepareLoadByHandle(
+		    bleAdvHandle, GAP_ADV_FREE_OPTION_DONT_FREE) != SUCCESS) {
 		Log_printf(LogModule_Beacon, Log_ERROR,
-			   "Failed to stop advertising");
+			   "Failed to prepare adv data load");
 		return;
 	}
 
-	if (BLEAppUtil_initAdvSet(&bleAdvHandle, &hubbleInitAdvSet) != SUCCESS) {
+	if (GapAdv_loadByHandle(bleAdvHandle, GAP_ADV_DATA_TYPE_ADV,
+				sizeof(advData), advData) != SUCCESS) {
 		Log_printf(LogModule_Beacon, Log_ERROR,
-			   "Failed to initialize advertising set");
+			   "Failed to load adv data");
 		return;
 	}
-
-	(void)BLEAppUtil_advStart(bleAdvHandle, &hubbleStartAdvSet);
 
 	Log_printf(LogModule_Beacon, Log_INFO, "Advertise data updated");
 }


HongNguyen635 (approved):
I let this run for some time and looks like this fixed the issue

PR #209, opened 2026-05-28
ESP-IDF: add Satellite support for ESP32C6

HongNguyen635:
Add Satellite support for ESP32C6 along with a quickstart on how to download Espressif PHY Lib early access blob.

### DIFF
--- a/docs/quickstart/esp-idf/index.rst
+++ b/docs/quickstart/esp-idf/index.rst
@@ -1,0 +1,104 @@
+.. _esp_idf_quick_start:
+
+ESP-IDF Quick Start
+===================
+
+This guide explains how to integrate the Hubble Network SDK into an
+`ESP-IDF <https://docs.espressif.com/projects/esp-idf/en/latest/esp32/get-started/index.html>`_
+project. The SDK ships as an ESP-IDF component that can be pulled into any
+ESP-IDF application via ``EXTRA_COMPONENT_DIRS``.
+
+
+Prerequisites
+*************
+
+- `ESP-IDF <https://docs.espressif.com/projects/esp-idf/en/latest/esp32/get-started/index.html>`_
+  installed and exported in the current shell (e.g. ``source ~/esp/esp-idf/export.sh``).
+- A supported ESP32 target. The BLE Network module works on any ESP32 chip
+  with a Bluetooth® Low Energy controller supported by ESP-IDF. The Satellite
+  Network module currently targets the **ESP32-C6**.
+- The Hubble Network SDK cloned or added to your project as a Git submodule.
+
+
+Adding Hubble Network to an ESP-IDF Project
+*******************************************
+
+#. Add the Hubble Network SDK to your application, for example as a submodule:
+
+   .. code-block:: bash
+
+      git submodule add https://github.com/HubbleNetwork/hubble-device-sdk
+
+#. Point your top-level ``CMakeLists.txt`` at the SDK's ESP-IDF component
+   directory using ``EXTRA_COMPONENT_DIRS``, and add ``hubblenetwork-sdk`` to
+   the component list:
+
+   .. code-block:: cmake
+
+      cmake_minimum_required(VERSION 3.13.1)
+
+      set(EXTRA_COMPONENT_DIRS path/to/hubble-device-sdk/port/esp-idf/)
+
+      set(COMPONENTS
+              "main"
+              "bt"
+              "hubblenetwork-sdk"
+      )
+
+      include($ENV{IDF_PATH}/tools/cmake/project.cmake)
+
+      project(my-app LANGUAGES C)
+
+#. Enable the desired Hubble modules. Either run ``idf.py menuconfig`` and
+   toggle them under *Component config* → *Hubble*, or set them in your
+   project's ``sdkconfig.defaults``:
+
+   .. code-block:: kconfig
+
+      # Terrestrial (BLE) Network
+      CONFIG_HUBBLE_BLE_NETWORK=y
+
+      # Satellite Network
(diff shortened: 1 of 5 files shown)


port/esp-idf/hubblenetwork-sdk/boards/esp32c6/radio.c
```diff
@@ -0,0 +1,222 @@
...
+
+	esp_phy_rftest_init();
+	return 0;
+}
+
+int hubble_sat_board_enable(void)
+{
+	esp_phy_rftest_config(1);
+	phy_set_step_01k(true);
+	return _timer_enable();
```
ceolin: you have to check the return here and in case of failure undo the previous instructions like (as it is done in `hubble_sat_board_disable`)

port/esp-idf/hubblenetwork-sdk/boards/esp32c6/radio.c
```diff
@@ -0,0 +1,222 @@
...
+	int ret = 0;
+
+	ret = _timer_init();
+	if (ret != 0) {
+		return ret;
+	}
+
+	_symbol_sem = xSemaphoreCreateBinary();
+	if (_symbol_sem == NULL) {
+		return -ENOMEM;
```
ceolin, line 224: Shouldn't you destroy / de-allocate the timer was created in `_timer_init()` with `gptimer_new_timer` ?

port/esp-idf/hubblenetwork-sdk/boards/esp32c6/radio.c
```diff
@@ -0,0 +1,222 @@
...
+	};
+
+	err = gptimer_set_raw_count(_timer_handle, 0);
+	if (err != ESP_OK) {
+		return err;
+	}
+
+	err = gptimer_set_alarm_action(_timer_handle, &alarm_config);
+	if (err != ESP_OK) {
+		return err;
```
ceolin: Not blocker but we could have some standard errno errors returned here, or at least check that we are returning negative values, because the majority of our APIs return negative values in case of error.

This can be addressed in a following commit though.

port/esp-idf/hubblenetwork-sdk/boards/esp32c6/radio.c
```diff
@@ -0,0 +1,222 @@
...
+	ret = gptimer_new_timer(&_timer_config, &_timer_handle);
+	if (ret != ESP_OK) {
+		return ret;
+	}
+
+	gptimer_event_callbacks_t cb = {
+		.on_alarm = _timer_cb,
+	};
+
+	return gptimer_register_event_callbacks(_timer_handle, &cb, NULL);
```
ceolin: is that ok, passing a variable allocated in the stack ? Does `gptimer_register_event_callbacks` copies the callback ?

port/esp-idf/hubblenetwork-sdk/boards/esp32c6/radio.c
```diff
@@ -0,0 +1,222 @@
...
+
+	phy_tx_tone(false, true, DEFAULT_TX_POWER_DBM);
+
+	(void)_timer_start(delay);
+	xSemaphoreTake(_symbol_sem, portMAX_DELAY);
+}
+
+int hubble_sat_board_init(void)
+{
+	int ret = 0;
```
ceolin: You don't need to initialize it since it here, there is no code path that leads to use this variable without it be initialized.

docs/quickstart/esp-idf/index.rst
```diff
@@ -0,0 +1,107 @@
...
+
+   `libphy_C6_20260317_c83212e.zip <https://dl.espressif.com/AE/libphy_C6_20260317_c83212e%20(2).zip>`_
+
+#. Unzip the archive and copy the extracted ``*.a`` files over the matching
+   files in your ESP-IDF installation:
+
+   .. code-block:: bash
+
+      unzip "libphy_C6_20260317_c83212e.zip"
+      cp *.a $HOME/esp/esp-idf/components/esp_phy/lib/esp32c6/
```
ceolin: instead of `$HOME/...` use `$ENV{IDF_PATH}`. It is consistent with the rest of the document and more precise.

docs/quickstart/esp-idf/index.rst
```diff
@@ -0,0 +1,107 @@
...
+
+   Replace ``$HOME/esp/esp-idf`` with the actual path to your ESP-IDF
+   installation if it lives elsewhere.
```
ceolin: Then you don't need it.

ceolin (changes requested):
It is looking good, a few comments only.

port/esp-idf/hubblenetwork-sdk/boards/esp32c6/radio.c
HongNguyen635: that's a good point, I followed espressif example, and yes, the callback gets copied inside `gptimer_register_event_callbacks` [here](https://github.com/espressif/esp-idf/blob/2cff0eae16e5d43779447c18c29f460d9af3183b/components/esp_driver_gptimer/src/gptimer.c#L298). So it copies the `_timer_cb` function pointer, not storing a reference to the `gptimer_event_callbacks_t struct`.

port/esp-idf/hubblenetwork-sdk/boards/esp32c6/radio.c
HongNguyen635: Thanks, I decided to add the esp err -> errno conversion in the same commit.

ceolin (approved):

PR #210, opened 2026-05-28
sat: zephyr: implement DTM API and add DTM sample

HongNguyen635:
Implement DTM API for Nordic and SiLabs platforms, along with adding another sample for Sat DTM.


port/zephyr/boards/nordic/nrf52/nrf21540dk.c
HongNguyen635, line 1: I'm just very conflicted on doing it this way and not sure if adding the 
`int hubble_sat_soc_power_set(int8_t in_power, nrf_radio_txpower_t *out);` in the `sat_soc.h` is the right way to do this.

IMO it is better to let the SoC deal with its SoC power settings, so:
- SoC: set SoC power.
- boards: set its board power (e.g prepare / set external PA) and then call the SoC to set the SoC power).

This way, there won't be a duplicated of the switch statement across all different boards. But I just don't like how the SoC power function signature is looking right now, and not sure if there is another way to do this.

ceolin: I don't have a very hard opinion about it since DTM allow fine control over the power but the reason I haven't added it to SoC is because ultimately the board is the one that
have enough information about the power to set. The PA has to be accounted when setting the power and the board is the only one that can do it. Regarding the duplicated
code, I don't know how much you can avoid since you can't let bypass the value direct to SoC, you need to check it and make the decision wether or not you need to turn the PA.

It seems to me that there is no code saving in most situations.

port/zephyr/boards/nordic/nrf52/nrf52_soc.c
```diff
@@ -188,3 +188,73 @@ int hubble_sat_soc_packet_send(const struct hubble_sat_packet_frames *packet)
 
 	return 0;
 }
+
+#ifdef CONFIG_HUBBLE_SAT_NETWORK_DTM_MODE
+
+int hubble_sat_soc_power_set(int8_t in_power, nrf_radio_txpower_t *out)
```
ceolin: You don't really need `out` here do you ? The caller knows what it requested and in case it is an invalid value it checks the return. Also it is weird that you have this function in lowest abstraction layer but that does not really sets the power, the board still needs to call `nrf_radio_txpower_set`

port/zephyr/boards/nordic/nrf52/nrf52_soc.c
```diff
@@ -188,3 +188,73 @@ int hubble_sat_soc_packet_send(const struct hubble_sat_packet_frames *packet)
...
+	switch (in_power) {
+	case 0:
+	case 2:
+	case 3:
+	case 4:
+	case 5:
+	case 6:
+	case 7:
+	case 8:
+		*out = in_power;
```
ceolin: This conversation is danger, if they change the value of their constants it will break silence, you should explicitly convert `in_power` to the appropriate constant, not only the negatives.

port/zephyr/boards/nordic/nrf52/nrf52_soc.c
```diff
@@ -188,3 +188,73 @@ int hubble_sat_soc_packet_send(const struct hubble_sat_packet_frames *packet)
 
 	return 0;
 }
+
+#ifdef CONFIG_HUBBLE_SAT_NETWORK_DTM_MODE
+
+int hubble_sat_soc_power_set(int8_t in_power, nrf_radio_txpower_t *out)
+{
+	switch (in_power) {
+	case 0:
```
ceolin: several code guidelines disallow this silence fall through and request an explicit assignment or comment acknowledging you are falling through.

port/zephyr/boards/nordic/nrf52/nrf52_soc.c
```diff
@@ -188,3 +188,73 @@ int hubble_sat_soc_packet_send(const struct hubble_sat_packet_frames *packet)
...
+		*out = NRF_RADIO_TXPOWER_NEG8DBM;
+		break;
+	case -4:
+		*out = NRF_RADIO_TXPOWER_NEG4DBM;
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	return 0;
```
ceolin, line 252: I think you should set the power here, it does not make much sense the API be asked to set the power and don't do it.

port/zephyr/boards/nordic/nrf53/nrf53_soc.c
```diff
@@ -230,3 +235,97 @@ int hubble_sat_soc_packet_send(const struct hubble_sat_packet_frames *packet)
...
+	default:
+		return -EINVAL;
+	}
+
+	if (in_power > 0) {
+		/* High voltage increase the output power by 3 dB */
+		*out -= 3;
+		nrf_vreqctrl_radio_high_voltage_set(NRF_VREQCTRL, true);
+	} else {
+		nrf_vreqctrl_radio_high_voltage_set(NRF_VREQCTRL, false);
```
ceolin, line 324: this just emphasizes the point about setting the power in this API.

samples/zephyr/sat-dtm/sample.yaml
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2026 Hubble Network, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+sample:
+  description: Hubble Network Satellite DTM example
+  name: hubble-sat-dtm
+common:
+  build_only: true
```
ceolin, line 8: if it is build only you should build for the targets that have it implemented as well.

samples/zephyr/sat-dtm/src/main.c
```diff
@@ -0,0 +1,442 @@
...
+#include <hubble/sat/packet.h>
+
+#define THREAD_TRANSMIT_PRIORITY K_LOWEST_APPLICATION_THREAD_PRIO
+#define STACKSIZE                2048
+
+#define HUBBLE_NUM_CHANNELS      (19)
+#ifdef CONFIG_SOC_SERIES_NRF52
+#define HUBBLE_RADIO_POWER_OPTIONS                                             \
+	"8, 7, 6, 5, 4, 3, 2, 0, -4, -8, -12, -16, -20, -40"
+#elif CONFIG_SOC_COMPATIBLE_NRF5340_CPUNET
```
ceolin: these macros does not scale for a generic app ... I think it is more appropriate assume the user must know what are the available powers for their target, specially because this is is aim for people that will go through a certification process, they will know this.

samples/zephyr/sat-dtm/src/main.c
```diff
@@ -0,0 +1,442 @@
...
+			      0, K_FOREVER);
+	if (tid == NULL) {
+		goto err;
+	}
+
+	_transmission_in_progress = true;
+	ret = true;
+	k_thread_start(tid);
+
+err:
```
ceolin: Maybe `end` is a better name for this label since it covers not only the error path.

samples/zephyr/sat-dtm/src/main.c
```diff
@@ -0,0 +1,442 @@
...
+error:
+	shell_print(sh, "Failed to set payload len. Valid options are: "
+			"-1, 0, 4, 9 or 13\n");
+	return -1;
+}
+
+static int cmd_transmit(const struct shell *sh, size_t argc, char **argv)
+{
+	ARG_UNUSED(argc);
+	ARG_UNUSED(argv);
```
ceolin, line 269: `sh` is also not used.

samples/zephyr/sat-dtm/src/main.c
```diff
@@ -0,0 +1,442 @@
...
+		goto error;
+		break;
+	}
+
+	return 0;
+
+error:
+	shell_print(sh, "Failed to set payload len. Valid options are: "
+			"-1, 0, 4, 9 or 13\n");
+	return -1;
```
ceolin: `-EINVAL` is more appropriate here.

ceolin (changes requested):
A few comments and it is looking nice :)

port/zephyr/boards/nordic/nrf52/nrf52_soc.c
HongNguyen635: this is also where I don't like how I am doing with the setting power. Because currently, the board also sets the SoC power, so I don't know if I should leave it as is or just moves all the SoC power setting to the SoC.

I've changed the implementation around a bit, will push the commits soon for you to check if that makes more sense.

samples/zephyr/sat-dtm/sample.yaml
HongNguyen635: I encountered the issue that I needs to fetch the blobs for it to work 😞 
so silabs needs the RAIL blob, nordic needs the Hubble SDK blob...

ceolin (approved):

PR #211, opened 2026-05-28
sat: pass: API refinements 

ceolin:
- Add max elevation angle parameter in the pass info
- Remove ascending from pass info

Look individual commit messages for more details.

### DIFF
--- a/src/hubble_sat_ephemeris.c
+++ b/src/hubble_sat_ephemeris.c
@@ -558,6 +558,26 @@ static double _lon_tolerance_get(double lat, double sat_altitude)
 			      (earth.radius * _cos(_DEG2RAD(lat)))));
 }
 
+static double _elevation_angle_get(double sat_lon, double dev_lat,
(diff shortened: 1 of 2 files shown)


src/hubble_sat_ephemeris.c
```diff
@@ -1001,8 +1038,7 @@ int hubble_next_pass_region_get(uint64_t t,
 			 "ascending=%d",
 			 (unsigned long long)pass->start,
 			 (unsigned long long)pass->culmination, pass->lon,
-			 (unsigned long long)pass->duration,
-			 (int)pass->ascending);
+			 (unsigned long long)pass->duration, (int)ascending);
```
HongNguyen635: should we include the elevation angle in all of the DEBUG message as well?

HongNguyen635 (commented):
Minor comment, otherwise I cross-checked the max elevation angle with the backend and they looks correct.

HongNguyen635 (approved):

PR #214, opened 2026-06-01
docs: Group ephemeris APIs so they render in published docs

buckleypaul:
## Problem

The satellite ephemeris APIs (`hubble_sat_satellites_set`, `hubble_next_pass_get`, `hubble_next_pass_region_get`) and their structs are fully documented in `include/hubble/sat/ephemeris.h`, but they do **not** appear in the auto-generated GitHub Pages documentation.

## Root cause

The docs are built with Sphinx + Breathe. The satellite API page (`docs/satellite/api.rst`) pulls symbols in **only** via Breathe `.. doxygengroup::` directives, which render only symbols that belong to a named Doxygen group.

- `sat.h` → `@defgroup hubble_sat_api` ✅ rendered
- `packet.h` → `@defgroup hubble_sat_packet_api @ingroup hubble_sat_api` ✅ rendered
- `ephemeris.h` → every symbol had `@brief` but **none belonged to a group** ❌ → Doxygen scanned them, but Breathe had nothing to emit

## Fix

Wrap the ephemeris structs and functions in `@addtogroup hubble_sat_api { ... }`, adding them to the existing satellite API group already referenced by `api.rst`. No `.rst` change required.

## Verification

CI's `build-docs` workflow (or local `make doxygen && make html` in `docs/`) should now render the ephemeris structs and functions on the satellite API page.

## Note

`include/hubble/sat/dtm.h` has the same "no group" issue and is also absent from the docs. Left out of scope here — can be folded in if desired.

### DIFF
--- a/include/hubble/sat/ephemeris.h
+++ b/include/hubble/sat/ephemeris.h
@@ -21,6 +21,11 @@ extern "C" {
 #endif
 
 /**
+ * @addtogroup hubble_sat_api
+ * @{
+ */
+
+/**
  * @struct hubble_sat_orbital_params
  * @brief Represents the orbital parameters of a satellite.
  *
@@ -155,6 +160,9 @@ int hubble_next_pass_get(uint64_t t, const struct hubble_sat_device_pos *pos,
 int hubble_next_pass_region_get(uint64_t t,
 				const struct hubble_sat_device_region *region,
 				struct hubble_sat_pass_info *pass);
+
+/** @} */
+
 #ifdef __cplusplus
 }
 #endif


include/hubble/sat/ephemeris.h
```diff
@@ -20,6 +20,11 @@
 extern "C" {
 #endif
 
+/**
+ * @addtogroup hubble_sat_api
+ * @{
```
HongNguyen635, line 25: Should we add the `@brief` like other file and put it in its own group `@defgroup`? e.g.

```
/**
 * @brief Hubble Sat Network Ephemeris APIs
 * @defgroup hubble_sat_ephemeris_api Satellite Network ephemeris APIs
 * @ingroup hubble_sat_api 
 * @{
 */
```

ceolin: Not necessarily needed imho, group this api as long with the satellite is good enough.

That's said not against it.

ceolin (approved):

include/hubble/sat/ephemeris.h
buckleypaul: I thought about that, but really there isn't a situation where you'd use satellite without this for the most part... so really makes sense to just bundle it all together.

PR #215, opened 2026-06-01
samples: esp-idf: add sat-continuous sample

HongNguyen635:
Add sat-continous sample for ESP32-C6 with option to sync time with the Hubble Connect App.

### DIFF
--- a/.github/workflows/build-samples-esp-idf.yaml
+++ b/.github/workflows/build-samples-esp-idf.yaml
@@ -46,10 +46,19 @@ jobs:
           command: |
             set -euxo pipefail
             pip install idf-build-apps
-            idf-build-apps build -p samples/esp-idf --recursive --target ${{ matrix.espidf_target }}
+            idf-build-apps build \
+              -p samples/esp-idf \
+              --recursive \
+              --target ${{ matrix.espidf_target }} \
+              --manifest-file tools/ci/esp-idf-build-manifest.yaml
             for app in samples/esp-idf/*/; do
               app_name=$(basename "$app")
               target=${{ matrix.espidf_target }}
+              # Check and skip samples not enabled for this target by the manifest
+              if [ ! -d "$app/build" ]; then
+                echo "Skipping $app_name for $target (not enabled in manifest)"
+                continue
+              fi
               (cd "$app" && \
                  idf.py size-components --format json2 \
                    --output-file "size-components-${app_name}-${target}.json" && \
--- a/samples/esp-idf/sat-continuous/main/main.c
+++ b/samples/esp-idf/sat-continuous/main/main.c
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2026 Hubble Network, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+
+#include "esp_log.h"
+#include "nvs_flash.h"
+#include "esp_system.h"
+
+#include "mbedtls/base64.h"
+
+#include <hubble/hubble.h>
+#include <hubble/sat/packet.h>
+
+#ifdef CONFIG_SAMPLE_SYNC_TIME
+#include "time_sync.h"
+#endif
+
+#define SAT_TX_SLEEP_MS 2000
+
+static const char *APP_TAG = "sat_continuous";
+
+static uint64_t _unix_time_ms = 0xdeadbeef;
+static uint8_t _hubble_key[CONFIG_HUBBLE_KEY_SIZE];
+
+void app_main(void)
+{
+	esp_err_t err = 0;
+	struct hubble_sat_packet pkt;
+
+	/* NVS flash init, dependency of ble stack to store configs */
+	err = nvs_flash_init();
+	if (err == ESP_ERR_NVS_NO_FREE_PAGES ||
+	    err == ESP_ERR_NVS_NEW_VERSION_FOUND) {
+		ESP_ERROR_CHECK(nvs_flash_erase());
+		err = nvs_flash_init();
+	}
+	ESP_ERROR_CHECK(err);
+
+	/* Decode device key */
+	if (strlen(CONFIG_HUBBLE_DEVICE_KEY) != 0) {
+		size_t outlen = 0;
+		int err = mbedtls_base64_decode(
+			_hubble_key, sizeof(_hubble_key), &outlen,
+			(const unsigned char *)CONFIG_HUBBLE_DEVICE_KEY,
+			strlen(CONFIG_HUBBLE_DEVICE_KEY));
+
+		if (err != 0) {
+			ESP_LOGE(APP_TAG, "Invalid key provided!");
+			return;
+		}
+
+		ESP_LOGD(APP_TAG, "Device key decoded (%zu bytes):", outlen);
+		ESP_LOG_BUFFER_HEX_LEVEL(APP_TAG, _hubble_key, outlen,
+					 ESP_LOG_DEBUG);
+	}
+
+#ifdef CONFIG_SAMPLE_SYNC_TIME
+	err = ble_sync_time(&_unix_time_ms);
+	if (err != ESP_OK) {
+		ESP_LOGE(APP_TAG, "Failed to sync time, error: %d", err);
+		return;
+	}
+#endif
+
+	err = hubble_init(_unix_time_ms, _hubble_key);
+	if (err != 0) {
+		ESP_LOGE(APP_TAG,
+			 "Failed to initialize Hubble Sat Network, error: %d",
+			 err);
+		return;
+	}
+
+	ESP_LOGI(APP_TAG, "Starting Sat Transmission");
+
+	for (;;) {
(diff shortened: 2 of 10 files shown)


samples/esp-idf/sat-continuous/sdkconfig.defaults
```diff
@@ -0,0 +1,16 @@
...
+CONFIG_BT_NIMBLE_ENABLED=y
+CONFIG_BT_NIMBLE_50_FEATURE_SUPPORT=n
+
+# Sat dependencies
+CONFIG_ESP_PHY_ENABLE_CERT_TEST=y
+CONFIG_HUBBLE_SAT_NETWORK=y
+
+# Disable nonce check since continuous transmit
+# will exceed the daily limit, applicable for
+# testing purpose. This should be turn on for production.
```
ceolin: `should be turned on` ?

samples/esp-idf/sat-continuous/main/main.c
```diff
@@ -0,0 +1,103 @@
...
+
+static uint64_t _unix_time_ms;
+static uint8_t _hubble_key[CONFIG_HUBBLE_KEY_SIZE];
+
+void app_main(void)
+{
+	esp_err_t err = 0;
+	struct hubble_sat_packet pkt;
+
+	/* NVS flash init, depenency of ble stack to store configs */
```
ceolin: typo, `depenency`

samples/esp-idf/sat-continuous/main/main.c
```diff
@@ -0,0 +1,103 @@
...
+			_hubble_key, sizeof(_hubble_key), &outlen,
+			(const unsigned char *)CONFIG_HUBBLE_DEVICE_KEY,
+			strlen(CONFIG_HUBBLE_DEVICE_KEY));
+
+		if (err != 0) {
+			ESP_LOGE(APP_TAG, "Invalid key provided!");
+			return;
+		}
+
+		ESP_LOGD(APP_TAG, "Device key decoded (%d bytes):", outlen);
```
ceolin: `outlen` is unsigned, you should probably use `%zu` instead of `%d`

samples/esp-idf/sat-continuous/main/main.c
```diff
@@ -0,0 +1,103 @@
...
+#include "nvs_flash.h"
+#include "esp_system.h"
+
+#include "mbedtls/base64.h"
+
+#include <hubble/hubble.h>
+#include <hubble/sat/packet.h>
+
+#ifdef CONFIG_SAMPLE_SYNC_TIME
+#include "time_sync.c"
```
ceolin: Isn't this file compiled in a a different object unit ? This shouldn't work.

.github/workflows/build-samples-esp-idf.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
           command: |
             set -euxo pipefail
             pip install idf-build-apps
-            idf-build-apps build -p samples/esp-idf --recursive --target ${{ matrix.espidf_target }}
+            idf-build-apps build -p samples/esp-idf --recursive --target ${{ matrix.espidf_target }} --exclude samples/esp-idf/sat-continuous
```
ceolin: Do you really need to exclude it ? Can it be built even not using the proper `libphy` ?

ceolin (changes requested):
A few items but it is looking good.

samples/esp-idf/sat-continuous/main/main.c
HongNguyen635: Let me add a header file for this. Was thinking it's not needed.
I tetsed and it worked but I guess the proper way is the header file.

.github/workflows/build-samples-esp-idf.yaml
HongNguyen635: you are right. I initially thought the build failed because of the EA, but it's because of the boards matrix I passed in. I found a better way for this using manifest file, will update this.

samples/esp-idf/sat-continuous/main/main.c
```diff
@@ -0,0 +1,103 @@
...
+
+#ifdef CONFIG_SAMPLE_SYNC_TIME
+	err = ble_sync_time(&_unix_time_ms);
+	if (err != ESP_OK) {
+		ESP_LOGE(APP_TAG, "Failed to sync time, error: %d", err);
+		return;
+	}
+#else
+	/* Dummy value if time sync is disabled */
+	_unix_time_ms = 1U;
```
ceolin: I think you can simplify it initializing the variable in its declaration to avoid this condition. In case `CONFIG_SAMPLE_SYNC_TIME` is enabled the default value should not matter.

samples/esp-idf/sat-continuous/main/time_sync.c
```diff
@@ -0,0 +1,292 @@
...
+
+/*
+ * @brief Sync time with Hubble Connect App (BLE).
+ *
+ * @param epoch_ms Pointer to store the synced epoch time in milliseconds.
+ * @return ESP_OK on success, or an error code on failure.
+ */
+esp_err_t ble_sync_time(uint64_t *epoch_ms)
+{
+	esp_err_t err = 0;
```
ceolin: It does not need to be initialized and if so would be better to use a proper value like `ESP_OK`.

samples/esp-idf/sat-continuous/main/time_sync.c
```diff
@@ -0,0 +1,292 @@
...
+		return err;
+	}
+
+	/* Set host callbacks */
+	ble_hs_cfg.reset_cb = _on_stack_reset;
+	ble_hs_cfg.sync_cb = _on_stack_sync;
+	ble_hs_cfg.store_status_cb = ble_store_util_status_rr;
+
+	ble_store_config_init();
+	xTaskCreate(_nimble_host_task, "NimBLE Host", 4 * 1024, NULL, 5, NULL);
```
ceolin: Stack size and priority should be set in some `define` symbol for clarity.

samples/esp-idf/sat-continuous/sdkconfig.defaults
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2026 Hubble Network, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+# Bluetooth
+CONFIG_BT_ENABLED=y
+CONFIG_BT_NIMBLE_ENABLED=y
+CONFIG_BT_NIMBLE_50_FEATURE_SUPPORT=n
+
+# Sat dependencies
+CONFIG_ESP_PHY_ENABLE_CERT_TEST=y
```
ceolin: Is this needed ? If so should be included by our SDK automatically ?

samples/esp-idf/sat-continuous/main/time_sync.c
```diff
@@ -0,0 +1,292 @@
...
+
+	err = ble_gatts_add_svcs(_gatt_svr_svcs);
+	if (err != ESP_OK) {
+		ESP_LOGE(BLE_TAG, "Failed to add GATT services, error: %d", err);
+		return err;
+	}
+
+	/* Set host callbacks */
+	ble_hs_cfg.reset_cb = _on_stack_reset;
+	ble_hs_cfg.sync_cb = _on_stack_sync;
```
ceolin: Not a blocker at all, but I `_on_stack` namespace does not talk with me well, looking the name I would never guess it is Bluetooth related, maybe `_on_ble` or `_on_nimble` ... ?

samples/esp-idf/sat-continuous/main/time_sync.c
```diff
@@ -0,0 +1,292 @@
...
+	/* Register GATT services */
+	ble_svc_gap_init();
+	ble_svc_gatt_init();
+
+	/* Update GATT service counter and add the service */
+	err = ble_gatts_count_cfg(_gatt_svr_svcs);
+	if (err != ESP_OK) {
+		ESP_LOGE(BLE_TAG, "Failed to count GATT services, error: %d",
+			 err);
+		return err;
```
ceolin, line 258: There are a lot of cleanup that needs to be done in case of error, like `nimble_port_deinit`.

I don't really expect it to be super error resilient and sometimes adding a lot of error handling has the side effect of making hard to read the core part of the code, so if you find this is the case, maybe just adding a comment attesting it is fine for me.

ceolin (changes requested):
Just a few comments, some of them are ok to ignore if you want. Just check them and let me know please.

samples/esp-idf/sat-continuous/sdkconfig.defaults
HongNguyen635: yea that is needed. I forgot that we can do this in Kconfig, let me make another PR to address this

samples/esp-idf/sat-continuous/main/time_sync.c
HongNguyen635: good point. I just copied this over from esp gatt sample. Will change it to better names.

samples/esp-idf/sat-continuous/main/time_sync.c
HongNguyen635: That is fair. The reason why I don't do the deinit is when this return, `main` would return as well. And it makes the code a bit hard to read. I can put a comment explain this.

ceolin (approved):

PR #218, opened 2026-06-03
sat: ephemeris: pin n0 mean-motion defect against SGP4 truth

buckleypaul:
The ephemeris golden values are captured from the propagator's own output and re-baselined on every change, so they prove the predictor is stable, not correct. Validating instead against an independent SGP4 propagation of the same elements exposes a real defect: the fixture supplies n0 as the TLE/Kozai mean motion while the SDK consumes it as a nodal mean motion. J2 separates the two by ~3.6 s/orbit (~55 s/day), so the predicted culmination leaves the 160 s transmission window within ~2 days and the device transmits outside the real pass.

The added check holds four SGP4 reference culminations and is marked expected-fail, so the defect is pinned and visible without holding CI red; converting n0 to the nodal mean motion 2*pi/(mdot + argpdot) flips it green.

### DIFF
--- a/tests/zephyr/ephemeris/src/main.c
+++ b/tests/zephyr/ephemeris/src/main.c
@@ -135,6 +135,59 @@ ZTEST(satellite_ephemeris_test, test_satellite_ephemeris_calculation)
 	}
 }
 
+/*
+ * Independent-reference check: predicted culmination vs canonical SGP4 truth.
+ *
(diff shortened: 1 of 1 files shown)


ceolin:
Should we close it ? Our pass prediction is not based on sgp4 thus EPP parameters are different TLE parameters.

PR #220, opened 2026-06-03
sat: Name convention fixes

ceolin:
Cleans up the public naming of the satellite pass-prediction API so it
 follows the `hubble_sat_*` convention used across the rest of the
 satellite surface, and names the header after what it provides.


include/hubble/sat/pass_prediction.h
```diff
@@ -5,12 +5,12 @@
  */
 
 /**
- * @file ephemeris.h
- * @brief Hubble Network Satellite Ephemeris APIs
+ * @file pass_prediction.h
```
buckleypaul, line 8: Should you change the .c for this as well?

ceolin: yep ! completely missed it

src/hubble_sat_pass_prediction.c
```diff
@@ -14,7 +14,7 @@
 #define M_PI 3.14159265358979323846
 #endif
 
-#define HUBBLE_EARTH_RADIUS              6378136.999954619 /* Earth radius at equator */
+#define HUBBLE_EARTH_RADIUS 6378136.999954619 /* Earth radius at equator */
```
HongNguyen635, line 17: this is a really weird change from the formatter

HongNguyen635 (approved):

src/hubble_sat_pass_prediction.c
ceolin: it is ... I was debating about keeping it the way it was or just accept the linter change, ended up deciding to fight this battle :)

PR #222, opened 2026-06-04
sat: freertos/ti: implement DTM and add DTM sample

HongNguyen635:
- Add DTM config option.
- Implement DTM support for TI cc23 / cc27.
- Add DTM sample for TI cc27.

### DIFF
--- a/samples/freertos/ti/sat-dtm/src/hubble_ti_crypto.c
+++ b/samples/freertos/ti/sat-dtm/src/hubble_ti_crypto.c
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2026 HubbleNetwork
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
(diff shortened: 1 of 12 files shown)


samples/freertos/ti/sat-dtm/src/hubble_ti_crypto.c
```diff
@@ -0,0 +1,36 @@
+/*
```
ceolin, line 1: It is missing copyright and SPDX

ceolin (dismissed):
Looks good, just fix the header please

samples/freertos/ti/sat-dtm/src/hubble_ti_crypto.c
HongNguyen635: thanks, I added the copyright

ceolin (approved):

PR #223, opened 2026-06-05
sat: esp-idf: implement DTM API and add DTM sample

HongNguyen635:
- Implement DTM API for esp32-c6.
- Add DTM sample.


HongNguyen635:
will change the exclude esp-idf from build commit to using the manifest file method after #215 is merged.

ceolin (approved):

PR #227, opened 2026-06-08
sat: tdr: Change default TDR to 10 ppm

ceolin:
The default Time Drift Rate was 500 ppm, which far overestimates the drift of a typical crystal oscillator (usually in the 10-50 ppm range).

The inflated value caused the satellite stack to add far more drift compensation retransmissions than necessary, increasing airtime and power consumption for no reliability gain.


port/zephyr/Kconfig
HongNguyen635: change 500 here to 10 as well

ceolin: thanks for catching it

HongNguyen635 (approved):

PR #228, opened 2026-06-08
docs: sat: document behavior of transmission modes

ceolin:
Document the baseline retry count and retransmission interval for each hubble_sat_transmission_mode (NORMAL: 8 retries / 20s, HIGH: 16 retries / 10s). Also explain that extra retries are added to compensate for clock drift accumulated since the last time sync, estimated from the device's Time Drift Rate (CONFIG_HUBBLE_SAT_NETWORK_DEVICE_TDR).

### DIFF
--- a/include/hubble/sat.h
+++ b/include/hubble/sat.h
@@ -31,15 +31,37 @@ extern "C" {
  * @brief Satellite transmission mode
  *
  * It tells what is the desired reliability when transmitting
- * a packet. Higher reliability consumes higher power and takes
- * longer because it increases the number of retries.
+ * a packet. Higher reliability consumes higher power because it increases the
+ * number of retries.
+ *
+ * @note The retry counts listed below are baselines. Extra retries are
+ *       added to compensate for the device's clock drift accumulated since
+ *       the last time synchronization: one additional retry is added for
+ *       every full retransmission interval worth of drift. The longer the
+ *       device goes without synchronizing its clock, the more retries are
+ *       performed. Modes with no retries (@ref HUBBLE_SAT_RELIABILITY_NONE)
+ *       are not affected.
+ *
+ *       The drift is estimated from the device's Time Drift Rate (TDR),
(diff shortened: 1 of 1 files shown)


include/hubble/sat.h
```diff
@@ -31,15 +31,37 @@ extern "C" {
...
+ * a packet. Higher reliability consumes higher power because it increases the
+ * number of retries.
+ *
+ * @note The retry counts listed below are baselines. Extra retries are
+ *       added to compensate for the device's clock drift accumulated since
+ *       the last time synchronization: one additional retry is added for
+ *       every full retransmission interval worth of drift. The longer the
+ *       device goes without synchronizing its clock, the more retries are
+ *       performed. Modes with no retries (@ref HUBBLE_SAT_RELIABILITY_NONE)
+ *       are not affected.
```
HongNguyen635, line 43: might be unrelated but the pass prediction returns culmination along with pass start. But pass start has been accomodated with TDR and shift with other modes' retries interval. Should we note this point as well? That if they uses `HUBBLE_SAT_RELIABILITY_NONE`, they need to compute the start time or if they want to use pass->start, needs to remember that it has been shifted accordingly?

HongNguyen635 (approved):

include/hubble/sat.h
ceolin: I thought about it and I don't know exactly what is the best approach here. My approach is that if one is using `HUBBLE_SAT_RELIABILITY_NONE` they are almost on their own when comes to how to transmit. The pass prediction tries to accommodate the drift and that is kind of orthogonal to the number and frequency of retries.

ceolin: I will merge it since you are no blocking it as well, but feel free to iterate over it if you believe it makes more sense.

HongNguyen635: okay sounds good, I'll think about it later.

PR #230, opened 2026-06-08
sat: timeout: Add timeouts for sat transmission

ceolin:
Add timeouts to start transmissions and between symbols.

### DIFF
--- a/port/esp-idf/hubblenetwork-sdk/boards/esp32c6/radio.c
+++ b/port/esp-idf/hubblenetwork-sdk/boards/esp32c6/radio.c
@@ -37,6 +37,9 @@
 #define HUBBLE_BASE_FREQUENCY           2482U
 #define HUBBLE_CHANNEL_OFFSET(_channel) (((_channel) * 64) + 489)
 
+#define _TIME_US_TO_TICK(_time_us) ((_time_us / 1000U) / portTICK_PERIOD_MS)
+#define _TIME_S_TO_TICK(_time_s)   ((_time_s * 1000U) / portTICK_PERIOD_MS)
+
 /**
  * This semaphore is used to protect a packet transmission and avoid
  * race conditions.
@@ -195,17 +198,25 @@ static int _radio_cw_start(uint16_t step, uint32_t delay, uint32_t duration_us)
 		return ret;
 	}
 
-	xSemaphoreTake(_symbol_sem, portMAX_DELAY);
+	ret = !(xSemaphoreTake(_symbol_sem, _TIME_US_TO_TICK(2 * duration_us)) ==
+		pdTRUE);
 
 	/* Symbol off time */
 	phy_tx_tone(false, true, DEFAULT_TX_POWER_DBM);
+
+	if (ret != 0) {
+		return -ETIMEDOUT;
+	}
 
 	ret = _timer_start(delay);
 	if (ret != 0) {
 		return ret;
 	}
 
-	xSemaphoreTake(_symbol_sem, portMAX_DELAY);
+	if (xSemaphoreTake(_symbol_sem, _TIME_US_TO_TICK(2 * delay)) != pdTRUE) {
+		return -ETIMEDOUT;
+	}
+
 	return 0;
 }
 
@@ -273,7 +284,12 @@ int hubble_sat_board_packet_send(const struct hubble_sat_packet_frames *packet)
 	int ret = 0;
 	int8_t frame = -1;
 
-	xSemaphoreTake(_transmit_sem, portMAX_DELAY);
+	if (xSemaphoreTake(_transmit_sem,
+			   _TIME_S_TO_TICK(HUBBLE_SAT_TRANSMISSION_TIMEOUT_S)) !=
+	    pdTRUE) {
+		return -ETIMEDOUT;
+	}
+
 	xSemaphoreTake(_symbol_sem, 0); /* Reset semaphore */
 
 	for (uint8_t i = 0; i < packet->total_number_of_symbols; i++) {
--- a/port/freertos/boards/ti/cc23xx_cc27xx/radio.c
+++ b/port/freertos/boards/ti/cc23xx_cc27xx/radio.c
@@ -24,6 +24,7 @@
 /* Hubble */
 #include <hubble/port/sat_radio.h>
 #include <hubble/sat/packet.h>
+#include <hubble/port/sat_radio.h>
 
 /* DMM */
 #if defined(USE_DMM_OVRDE)
@@ -60,6 +61,8 @@
 #else
 #error "Device not supported"
 #endif
+
+#define _TIME_S_TO_TICK(_time_s) ((_time_s * 1000U) / portTICK_PERIOD_MS)
 
 /**
  * This semaphore is used to protect a packet transmission and avoid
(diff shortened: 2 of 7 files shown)


port/zephyr/boards/nordic/nrf52/nrf52_soc.c
```diff
@@ -18,11 +18,14 @@
...
+#define RADIO_NODE             DT_NODELABEL(radio)
 
 /* From NRF52840_PS_v1.2 6.20.15.8 Time between TXEN -> READ is 140us */
-#define WAIT_SYMBOL_OFF_US (HUBBLE_WAIT_SYMBOL_OFF_US - 140)
-#define WAIT_SYMBOL_US     (HUBBLE_WAIT_SYMBOL_US + 140)
+#define WAIT_SYMBOL_OFF_US     (HUBBLE_WAIT_SYMBOL_OFF_US - 140)
+#define WAIT_SYMBOL_US         (HUBBLE_WAIT_SYMBOL_US + 140)
+
+/* Max time for semaphore symbol to wait before failing. */
+#define WAIT_SYMBOL_TIMEOUT_US K_USEC(2 * (WAIT_SYMBOL_OFF_US + WAIT_SYMBOL_US))
```
HongNguyen635, line 28: is the 2 here arbitrary? Or is it related to the 2 uses in `HUBBLE_SAT_TRANSMISSION_TIMEOUT_S`?

port/zephyr/boards/nordic/nrf54/nrf54_soc.c
```diff
@@ -27,13 +27,16 @@
...
-#define WAIT_SYMBOL_US             (HUBBLE_WAIT_SYMBOL_US + 40)
+#define WAIT_SYMBOL_OFF_US     (HUBBLE_WAIT_SYMBOL_OFF_US - 40)
+#define WAIT_SYMBOL_US         (HUBBLE_WAIT_SYMBOL_US + 40)
 
-#define NRF_DPPIC                  NRF_DPPIC10
+/* Max time for semaphore symbol to wait before failing. */
+#define WAIT_SYMBOL_TIMEOUT_US K_USEC(2 * (WAIT_SYMBOL_OFF_US + WAIT_SYMBOL_US))
+
+#define NRF_DPPIC              NRF_DPPIC10
 #define RADIO_ENABLE_TX_ON_CC0_PPI 9U
```
HongNguyen635, line 40: another weird case from the formatter :(

port/zephyr/boards/silabs/efr32/radio.c
```diff
@@ -126,21 +127,24 @@ static int _radio_cw_start(uint8_t channel, uint16_t step, uint32_t delay,
...
 
-	k_sem_take(&_symbol_sem, K_FOREVER);
+	ret = k_sem_take(&_symbol_sem, (2 * duration_us));
 
 	status = sl_rail_stop_tx_stream(_rail_handle);
 	if (status != SL_RAIL_STATUS_NO_ERROR) {
 		return _sl_status_to_errno(status);
 	}
 
+	if (ret != 0) {
```
HongNguyen635, line 137: should we preserve the error from k_sem_take here?
Maybe:
```
return (ret != 0) ? ret : _sl_status_to_errno(status);
```
But I'm not sure if this makes it hard to read / maintain.

port/zephyr/boards/nordic/nrf52/nrf52_soc.c
HongNguyen635: I'm not sure how to feel about the use of constant `2` through the code. When we decide to change it to another number later, we'll have to go through the port and change each. Would it makes sense if we define something for it, e.g. `SYMBOL_TIMEOUT_FACTOR` or something like that...

port/freertos/boards/ti/cc23xx_cc27xx/radio.c
```diff
@@ -24,6 +24,7 @@
 /* Hubble */
 #include <hubble/port/sat_radio.h>
 #include <hubble/sat/packet.h>
+#include <hubble/port/sat_radio.h>
```
HongNguyen635, line 27: duplicate include?

port/freertos/boards/ti/cc23xx_cc27xx/radio.c
```diff
@@ -61,6 +62,8 @@
 #error "Device not supported"
 #endif
 
+#define _TIME_S_TO_TICK(_time_s) ((_time_s * 1000U) / portTICK_PERIOD_MS)
```
HongNguyen635, line 65: I'm a bit curious on is this preferred over `pd_MS_TO_TICKS( )`? I recall something about using `pd_MS_TO_TICKS` have some safety guard like promotion and type cast.

port/esp-idf/hubblenetwork-sdk/boards/esp32c6/radio.c
```diff
@@ -37,6 +37,9 @@
 #define HUBBLE_BASE_FREQUENCY           2482U
 #define HUBBLE_CHANNEL_OFFSET(_channel) (((_channel) * 64) + 489)
 
+#define _TIME_US_TO_TICK(_time_us) ((_time_us / 1000U) / portTICK_PERIOD_MS)
+#define _TIME_S_TO_TICK(_time_s)   ((_time_s * 1000U) / portTICK_PERIOD_MS)
```
HongNguyen635, line 41: I kind of feeling something off about this so I asked Claude and checked with ESP-IDF [source](https://github.com/espressif/esp-idf/blob/891341153d8051f0190e0b32caa038410aa9d532/components/freertos/FreeRTOS-Kernel/portable/linux/include/freertos/portmacro.h#L74):

```
#define portTICK_PERIOD_MS          ( ( TickType_t ) 1000 / configTICK_RATE_HZ )
```

-> `portTICK_PERIOD_MS` = 1000 / 100 = 10

where tick rate Hz is 100 by [default](https://github.com/espressif/esp-idf/blob/891341153d8051f0190e0b32caa038410aa9d532/components/freertos/Kconfig#L37). I also doubled check the `build` folder and the default of `CONFIG_FREERTOS_HZ` is indeed 100.

So say we have symbol off = 800 us -> (800 / 1000U) / 10 = 0
or symbol on = 8000 us -> ( 8000 / 1000 ) / 10 = 0

So either case it still fails.

port/zephyr/boards/nordic/nrf54/nrf54_soc.c
ceolin: yep, super annoying. I can try to revert it and see if CI complains

port/esp-idf/hubblenetwork-sdk/boards/esp32c6/radio.c
ceolin: Yep, good point, we should have at least 1 tick as min.

port/zephyr/boards/nordic/nrf52/nrf52_soc.c
ceolin: it is arbitrary, just two times the expected time, could be even less.

Let me re-struct it and make it consistent

ceolin:
Closing it since it is completely stale.

PR #234, opened 2026-06-10
feat: Add ble-beacon-mfg-data sample

buckleypaul:
Show SDK users how to attach their own data to a Hubble beacon without routing it through the encrypted Hubble payload. The sample emits a single advertisement carrying the Hubble service UUID, the Hubble service data, and a manufacturer-specific field (0xFCA6 prefix + a 4-byte counter incremented on each refresh) so the two coexist within the 31-byte legacy budget.

It uses the device uptime counter source so no UTC time has to be provisioned: hubble_init() takes 0 as the initial EID counter and it advances with uptime.

The master key is passed on the build command line as base64 (-DHUBBLE_KEY=...) and decoded to a byte array at build time (CMakeLists.txt invokes Python, already a Zephyr build requirement), so no base64 code ships on the device and there is no conf file or pre-build script to manage. A BUILD_ASSERT rejects a wrong-size key; omitting the key builds an all-zero placeholder for CI.

Kept deliberately bare-bones (no CTS sync, no NVS persistence) so it reads as reference material; the existing ble-beacon sample covers those.

### DIFF
--- a/samples/zephyr/ble-beacon-mfg-data/src/main.c
+++ b/samples/zephyr/ble-beacon-mfg-data/src/main.c
@@ -1,0 +1,189 @@
+/*
+ * Copyright (c) 2026 Hubble Network, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*
+ * Hubble Network BLE Beacon + Manufacturer Data sample.
+ *
+ * This sample shows how to emit a SINGLE BLE advertisement that contains both
+ * the standard Hubble beacon payload AND an application-owned
+ * "manufacturer specific data" field. The manufacturer field carries a 4-byte
+ * counter that is incremented every time the advertisement is refreshed
+ * (every CONFIG_HUBBLE_MFG_DATA_SAMPLE_UPDATE_PERIOD seconds, 5 minutes by
+ * default).
+ *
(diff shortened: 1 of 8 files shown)


samples/zephyr/ble-beacon-mfg-data/src/main.c
```diff
@@ -0,0 +1,189 @@
...
+ * The advertisement is built from three AD (Advertising Data) structures:
+ *
+ *   [1] BT_DATA_UUID16_ALL      -> 0xFCA6, the Hubble service UUID. This is how
+ *                                  Hubble gateways recognize the beacon.
+ *   [2] BT_DATA_SVC_DATA16      -> the encrypted Hubble payload produced by
+ *                                  hubble_ble_advertise_get(). The buffer
+ *                                  already starts with the 0xFCA6 service UUID.
+ *   [3] BT_DATA_MANUFACTURER_DATA -> application data. By BLE convention this
+ *                                  field starts with a 2-byte company
+ *                                  identifier; here we reuse Hubble's 0xFCA6
```
ceolin, line 25: Just noting that `0xFCA6` is a registered service UUID and not manufacturing id, but I think it is better to re-use it indeed.

ceolin (approved):

PR #235, opened 2026-06-10
feat: sat: derive packet keys from configured counter source

buckleypaul:
Satellite packet encryption always derived its time counter from Unix epoch time, ignoring CONFIG_HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME. Devices configured for uptime-based EID rotation (those without a UTC source) would therefore key satellite packets off uninitialized/zero Unix time, diverging from the BLE path and breaking key agreement with the backend.

Route the satellite path through hubble_counter_get() so it honors the same counter source and config values as BLE advertisements, giving both networks identical key-derivation behavior. The Unix-time-only helper is now unused and removed to keep the two paths from drifting again.


ceolin (approved):

src/hubble_sat_packet.c
HongNguyen635, line 304: would this change the payload protocol version?

I don't remember if the payload version is different for time and timeless, but worth a check

buckleypaul: No - it's a purely backend thing on which counter source we use. If it matches the backend, we're good.

HongNguyen635: okay that's good then.

HongNguyen635 (approved):

PR #236, opened 2026-06-10
api: sat: rename hubble_sat_packet_send to hubble_sat_broadcast

buckleypaul:
The name "send" implied a single, point-to-point transmission of one packet. In reality the call emits the packet repeatedly into a broadcast medium with no acknowledgement: depending on the reliability mode the same packet is transmitted up to 8 or 16 times (plus drift-compensation retries), and it is never directed at a specific receiver. "broadcast" sets the right expectation that this is a fire-and-forget, multi-shot operation that blocks for the whole transmission period, so callers budget power and timing accordingly rather than assuming a quick send.

The "packet" qualifier is dropped: it added no information the hubble_sat_packet argument type does not already convey, and the shorter hubble_sat_broadcast reads as the verb for the whole satellite channel.

### DIFF
--- a/docs/architecture/project-organization.rst
+++ b/docs/architecture/project-organization.rst
@@ -178,7 +178,7 @@ Use this namespace for stable application APIs and SDK-wide constants. Examples:
   ``hubble_key_set()`` from ``hubble.h``;
 * ``hubble_ble_advertise_get()`` and
   ``hubble_ble_advertise_expiration_get()`` from ``ble.h``;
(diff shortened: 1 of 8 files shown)


ceolin (changes requested):
No ... broadcast is a terrible name. "Broadcast" in RF/networking usually implies one-to-many, fire-and-forget — which misleads about both the directionality and the blocking, retried nature of the call.

Also dropping the packet

ceolin:
Closing since it is stale. Feel free to re-open if something changes.

PR #239, opened 2026-06-10
fixes: Random fixes / improvements

ceolin:
- Fix possible off by one issue in `_encode`. Although it is not a problem right now because of the size of buffers we use, the algorithm itself has an issue and may cause a problem in future.
- Seed PRNG in FreeRTOS to avoid possible same random values across boots

### DIFF
--- a/port/freertos/hubble_freertos.c
+++ b/port/freertos/hubble_freertos.c
@@ -15,6 +15,8 @@
 #endif
 
 #include <errno.h>
+#include <stdbool.h>
+#include <stdint.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


HongNguyen635 (approved):
Probably another PR but while re-reading the `hubble_bitarray`, I don't think I saw the [len](https://github.com/HubbleNetwork/hubble-device-sdk/blob/1410c148e56efd95ec1f9d187eb4e4f1c858292f/src/utils/bitarray.h#L19) field being used anywhere. Is this a remnant from an old design? If so, is it safe to remove then?

PR #240, opened 2026-06-12
samples: zephyr: Add sat-dual-stack sample

ceolin:
Add a sample that runs the Hubble Terrestrial (BLE) and Satellite networks together.

The application provisions UTC time and satellite orbital parameters over a BLE GATT service, schedules the next satellite pass using the SDK's pass-prediction APIs, advertises Hubble beacon packets while it waits, and transmits to the satellite when the pass timer expires.


samples/zephyr/sat-dual-stack/boards/nrf54l15dk_nrf54l15_cpuapp.conf
```diff
@@ -0,0 +1,7 @@
+CONFIG_NRFX_TIMER=y
+CONFIG_SAMPLE_PROVIDE_SAT_BOARD_SUPPORT=n
+CONFIG_BT_CTLR_ADVANCED_FEATURES=y
+CONFIG_BT_CTLR_DYNAMIC_INTERRUPTS=y
+CONFIG_DYNAMIC_INTERRUPTS=y
+CONFIG_DYNAMIC_DIRECT_INTERRUPTS=y
+CONFIG_SHARED_INTERRUPTS=y
```
HongNguyen635: just for style, can we add another line at the end?

samples/zephyr/sat-dual-stack/boards/xg24_rb4187c.conf
```diff
@@ -0,0 +1,2 @@
+CONFIG_SOC_GECKO_CUSTOM_RADIO_PHY=y
```
HongNguyen635, line 1: need to add this one for dual stack as well:
`CONFIG_SILABS_SISDK_RAIL_MULTIPROTOCOL=y`

samples/zephyr/sat-dual-stack/boards/xiao_mg24.conf
```diff
@@ -0,0 +1,2 @@
+CONFIG_SOC_GECKO_CUSTOM_RADIO_PHY=y
```
HongNguyen635, line 1: same thing here, add `CONFIG_SILABS_SISDK_RAIL_MULTIPROTOCOL=y`

samples/zephyr/sat-dual-stack/Kconfig
```diff
@@ -0,0 +1,33 @@
...
+    bool "Enable debug mode"
+    default n
+    help
+	Enable debug mode for the Hubble sample application.
+	This will set the sat timer for 2 minutes instead
+	of until a satellite pass is overhead.
+
+config SAMPLE_PROVIDE_SAT_BOARD_SUPPORT
+    bool "Provide satellite board suport for test purpose"
+    default y
```
HongNguyen635, line 22: we can have the default = n and have the `sample.yaml` y-selected it. Maybe that would make it less confusing on why we have this `CONFIG_SAMPLE_PROVIDE_SAT_BOARD_SUPPORT=n` on every supported boards.

samples/zephyr/sat-dual-stack/Kconfig
```diff
@@ -0,0 +1,33 @@
...
+    help
+	Targets that support satellite must select it, otherwise
+	board APIs will be mocked in the sample.
```
HongNguyen635: maybe rephrase the 1st clause as "Targets that support satellite must select and provide the board support", what do you think?

samples/zephyr/sat-dual-stack/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+This sample demonstrates running the **Hubble Terrestrial (BLE) Network** and the
+**Hubble Satellite Network** together on a single Zephyr device.
+
+## Overview
+
+The application:
+
+1. Provisions the device over a BLE GATT service (UTC time and satellite
+   orbital parameters).
```
HongNguyen635: and coordinate?

samples/zephyr/sat-dual-stack/README.md
```diff
@@ -0,0 +1,127 @@
...
+                       v
+       +-------------------------------+
+       |  Enable BLE, start            |
+       |  connectable provisioning adv |
+       +-------------------------------+
+                       |
+                       v
+       +-------------------------------+
+       |  GATT writes: UTC time +      |
+       |  orbital params, disconnect   |
```
HongNguyen635: same thing here, location as well?

samples/zephyr/sat-dual-stack/time-sync.py
```diff
@@ -0,0 +1,211 @@
...
+    device = await BleakScanner.find_device_by_filter(match_hubble_sync_uuid)
+    if device is not None:
+        async with BleakClient(device) as client:
+            data = bytearray([0x01, 0x02])
+            data += bytearray(
+                int(datetime.now(UTC).timestamp() * 1000).to_bytes(
+                    8, byteorder="little", signed=False
+                )
+            )
+            await client.write_gatt_char(_CHARACTERISTIC_UUID, data, response=True)
```
HongNguyen635: we probably should write time last, because the fw checks if time has written upon disconnect, so if orbital params or location failed midway, but time succeed, the sem is still given.

samples/zephyr/sat-dual-stack/time-sync.py
```diff
@@ -0,0 +1,211 @@
...
+    def match_hubble_sync_uuid(device: BLEDevice, adv: AdvertisementData):
+        for uuid in adv.service_uuids:
+            if uuid.startswith(HUBBLE_BLE_UUID_SYNC):
+                return True
+        return False
+
+    device = await BleakScanner.find_device_by_filter(match_hubble_sync_uuid)
+    if device is not None:
+        async with BleakClient(device) as client:
+            data = bytearray([0x01, 0x02])
```
HongNguyen635: should we pull out these 2 command header to a constants? May be better for readability and in case these change in the future.

samples/zephyr/sat-dual-stack/time-sync.py
```diff
@@ -0,0 +1,211 @@
...
+        level=logging.INFO,
+        format="%(asctime)s %(levelname)s %(message)s",
+    )
+    args = sys.argv[1:]
+
+    if len(args) == 2:
+        sync(float(args[0]), float(args[1]))
+    else:
+        # No lat/lon provided. Let's use Seattle
+        sync(lat=47.614376, lon=-122.319323)
```
HongNguyen635: can add a log indicate the default

samples/zephyr/sat-dual-stack/src/app_ble.c
```diff
@@ -0,0 +1,313 @@
...
+	k_timer_start(&_adv_refresh_timer, HUBBLE_ADV_REFRESH_PERIOD,
+		      HUBBLE_ADV_REFRESH_PERIOD);
+
+	return 0;
+}
+
+int ble_adv_stop(void)
+{
+	k_timer_stop(&_adv_refresh_timer);
+	bt_le_adv_stop();
```
HongNguyen635, line 160: cancel work here as well just in case?

samples/zephyr/sat-dual-stack/src/main.c
```diff
@@ -0,0 +1,192 @@
...
+
+#include <stdint.h>
+#include <string.h>
+
+#include "app_ble.h"
+
+LOG_MODULE_REGISTER(main, CONFIG_APP_LOG_LEVEL);
+
+#define HUBBLE_MAX_SAT 6U
+#define MS_PER_SEC     1000U
```
HongNguyen635: I think zephyr has this `MSEC_PER_SEC` so no need to redefine it.

HongNguyen635 (commented):
some comments + the reason why CI fail is because of the nrf52 blob we have for sat network.

samples/zephyr/sat-dual-stack/README.md
ceolin: Yep, I added the coordinates later and forgot updating it.

samples/zephyr/sat-dual-stack/time-sync.py
ceolin: I don't worry much about it, we can do whatever order we want in the companion but that is not enough because these interfaces don't enforce it. We can / should make the application more resilient but it makes it more complex and kind of mess with the idea of having easy and simple to understand.

samples/zephyr/sat-dual-stack/src/app_ble.c
ceolin: There is no much benefit in doing that imho. First, you can't simply cancelling, you have to cancel and check if it work is being processed and if so wait until it finishes.
The work here is just updating the advertisment data that checks if bluetooth is enabled,
my understanding is that even if it is fired it won't to anything.

samples/zephyr/sat-dual-stack/Kconfig
ceolin: It is trickier, because then you have to exclude the targets that implement it

samples/zephyr/sat-dual-stack/README.md
```diff
@@ -0,0 +1,127 @@
+# Hubble Satellite Dual-Stack on Zephyr
+
+This sample demonstrates running the **Hubble Terrestrial (BLE) Network** and the
+**Hubble Satellite Network** together on a single Zephyr device.
+
+## Overview
+
+The application:
+
+1. Provisions the device over a BLE GATT service (Unix ephoc time, satellite
```
HongNguyen635: typo: epoch

samples/zephyr/sat-dual-stack/README.md
```diff
@@ -0,0 +1,127 @@
...
+```sh
+west build -b <board> samples/zephyr/sat-dual-stack \
+    -- -DCONFIG_HUBBLE_DEVICE_KEY=\"<your-base64-key>\"
+west flash
+```
+
+## Provisioning
+
+On boot the device starts a connectable advertisement named **"Hubble-Zephyr"**
+(UUID `0xFCA7`) and waits for provisioning. A companion app, or provided script,
```
HongNguyen635, line 45: we should mentioned the steps about how to use the script here. e.g.

Install Python dependencies and run the *dual-stack-companion.py* provisioning script:

```bash
pip install -r ../../../../tools/requirements-companion.txt

export HUBBLE_API_TOKEN=<your-hubble-api-token>

python ../../../../tools/dual-stack-companion.py
```

HongNguyen635 (commented):
Some minor comments. I tested on SiLabs and it's working great.

samples/zephyr/sat-dual-stack/README.md
ceolin: I don't like it duplicated all over the place tbh, but for now lets do this.

HongNguyen635 (approved):

PR #241, opened 2026-06-12
sat: nrf54: Fix issues with dual-stack

ceolin:
There were some issues when switching from Bluetooth to satellite transmission depending on how the Bluetooth stack was used. If it was just beaconing before the been disabled it seems some events and/or states were not being properly cleared messing with the satellite transmissions.

This commit clears up all DPPIC radio and timer channels and shorts config.

### DIFF
--- a/port/zephyr/boards/nordic/nrf54/nrf54_soc.c
+++ b/port/zephyr/boards/nordic/nrf54/nrf54_soc.c
@@ -140,6 +140,66 @@ static void _timer_setup(void)
 	nrfx_timer_init(&_timer0, &timer_cfg, NULL);
 }
 
+/**
+ * Tear down any radio/timer wiring left behind by the BLE controller.
+ *
+ * The BLE controller drives the radio entirely through DPPI: it leaves the
+ * RADIO connected to DPPIC channels via its SUBSCRIBE and PUBLISH registers,
+ * keeps those DPPIC channels enabled, may leave the radio in a non-DISABLED
+ * state, and leaves stale shorts/config behind.
+ */
+static void _radio_reset(void)
+{
+	const nrf_radio_task_t tasks[] = {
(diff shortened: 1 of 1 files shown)


port/zephyr/boards/nordic/nrf54/nrf54_soc.c
```diff
@@ -140,6 +140,71 @@ static void _timer_setup(void)
...
+	nrf_timer_task_trigger(_timer0.p_reg, NRF_TIMER_TASK_STOP);
+	nrf_timer_int_disable(_timer0.p_reg, ~0U);
+	nrf_timer_shorts_set(_timer0.p_reg, 0);
+	nrf_timer_publish_clear(_timer0.p_reg, NRF_TIMER_EVENT_COMPARE0);
+	nrf_timer_publish_clear(_timer0.p_reg, NRF_TIMER_EVENT_COMPARE1);
+
+	/* Finally, disable every DPPIC channel so no other peripheral's stale
+	 * wiring can drive the radio or our timer mid-transmission.
+	 */
+	nrf_dppi_channels_disable_all(NRF_DPPIC);
```
HongNguyen635: I'm just wondering if this is safe to do this. Is dppi exclusive to the radio wiring? If this is not the case we could be disabling various other peripherals that uses it and the after sat finishes, other peripherals need to re-enable the wiring.

HongNguyen635 (dismissed):

port/zephyr/boards/nordic/nrf54/nrf54_soc.c
ceolin: Ok, I am convinced to remove it. I got afraid / worried with dangling events messing with radio and timer that ended up with this nuke. I should be fine to document that application should not mess with radio and the used timer during this period.
I have tested with bluetooth controller and it is ok without this.

HongNguyen635 (approved):

PR #242, opened 2026-06-12
board: nrf54: Add SoftDevice-compatible SoC backend

ceolin:
When the SoftDevice Controller (CONFIG_BT_LL_SOFTDEVICE) is enabled, MPSL owns the radio peripheral and it can only be accessed from within a granted timeslot. The existing nrf54_soc.c controls RADIO and TIMER0 directly, which conflicts with that model.

Add nrf54_soc_softdevice.c, an alternate SoC backend that transmits Hubble satellite packets inside MPSL timeslots.

### DIFF
--- a/port/zephyr/boards/nordic/nrf54/nrf54_soc_softdevice.c
+++ b/port/zephyr/boards/nordic/nrf54/nrf54_soc_softdevice.c
@@ -1,0 +1,377 @@
+/*
+ * Copyright (c) 2026 Hubble Network, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <sat_soc.h>
+
+#include <zephyr/kernel.h>
+#include <zephyr/irq.h>
+#include <zephyr/types.h>
+
+#include <hubble/port/sat_radio.h>
+#include <hubble/sat/packet.h>
+
+#include <hal/nrf_radio.h>
+#include <hal/nrf_dppi.h>
+#include <hal/nrf_timer.h>
+
+#include <mpsl_timeslot.h>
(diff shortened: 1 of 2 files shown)


ceolin:
~~Still WIP, I will common code to a common file.~~

port/zephyr/boards/nordic/nrf54/nrf54_soc_softdevice.c
```diff
@@ -0,0 +1,377 @@
...
+
+/**
+ * Signature for APIs provided by the binary library.
+ */
+int hubble_nrf_lib_frequency_set(uint8_t channel, uint8_t step);
+int hubble_nrf_lib_enable(void);
+int hubble_nrf_lib_disable(void);
+
+/* TX power applied at the start of every timeslot. */
+static nrf_radio_txpower_t _sat_power = RADIO_TXPOWER_TXPOWER_0dBm;
```
HongNguyen635: would this cause conflict if DTM is enabled? Since dtm has 
`extern nrf_radio_txpower_t _sat_power;`

HongNguyen635 (commented):
Do you think we should add Kconfig fragments for soft device for the samples as well?

port/zephyr/boards/nordic/nrf54/nrf54_soc_softdevice.c
ceolin: > Do you think we should add Kconfig fragments for soft device for the samples as well?

I am thinking the best way to do it, specially because we are looking to support different versions, this makes things more complicated.

HongNguyen635 (approved):

PR #243, opened 2026-06-15
sat: pass-prediction: add  minimum elevation angle setter

ceolin:
Replace the compile-time HUBBLE_ELEVATION_ANGLE_TOLERANCE macro with a runtime setter, hubble_sat_min_elevation_angle_set(), so applications can tune the minimum elevation that a pass must reach to be valid.

### DIFF
--- a/include/hubble/sat/pass_prediction.h
+++ b/include/hubble/sat/pass_prediction.h
@@ -161,6 +161,24 @@ int hubble_sat_next_pass_region_get(
 	uint64_t t, const struct hubble_sat_device_region *region,
 	struct hubble_sat_pass_info *pass);
 
+/**
+ * @brief Set the minimum elevation angle for a satellite pass.
+ *
+ * This function configures the minimum elevation angle above the horizon that a
+ * satellite must reach at culmination for a pass to be reported by
+ * hubble_sat_next_pass_get() and hubble_sat_next_pass_region_get(). Passes whose
+ * maximum elevation angle is below this threshold are ignored.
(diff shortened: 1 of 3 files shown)


include/hubble/sat/pass_prediction.h
```diff
@@ -161,6 +161,24 @@ int hubble_sat_next_pass_region_get(
...
+ * hubble_sat_next_pass_get() and hubble_sat_next_pass_region_get(). Passes whose
+ * maximum elevation angle is below this threshold are ignored.
+ *
+ * Raising the angle restricts results to higher, closer passes (typically better
+ * link quality) at the cost of fewer reported passes. The default minimum
+ * elevation angle is 45 degrees.
+ *
+ * @param angle Minimum elevation angle in degrees. Must be in the range [30, 90].
+ * @return 0 on success.
+ * @retval -EINVAL if @p angle is outside the range [30, 90].
```
HongNguyen635, line 178: non blocking but I thought we want [45, 90]

HongNguyen635 (approved):
a minor comment but not blocking.

include/hubble/sat/pass_prediction.h
ceolin: 45 is default

HongNguyen635: Yeah but I thought we aso want to limit just to anything above 45

Edit: clarified offline, we're good with [30, 90] then.

PR #244, opened 2026-06-15
samples: zephyr: sat-continuous: provision key at build time

buckleypaul:
Adds a real provisioning path to the `sat-continuous` sample, mirroring `ble-beacon-mfg-data`.

- **Build-time key**: pass the base64 master key via `-DHUBBLE_KEY`; decoded into a byte array at configure time (no base64 in firmware). All-zero placeholder + warning if omitted so CI still builds.
- **Encryption counter**: switch to the device uptime counter source (`CONFIG_HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME`), so no UTC provisioning needed.
- **Nonce check**: disabled for this sample (continuous TX, no NVS).
- **TX interval**: new `CONFIG_SAMPLE_SAT_TX_INTERVAL_SECONDS` Kconfig (default 5s), used in `main.c`.
- **README**: usage docs, including the `west blobs fetch hubblenetwork-sdk` step required for Nordic radios.

### DIFF
--- a/samples/zephyr/sat-continuous/CMakeLists.txt
+++ b/samples/zephyr/sat-continuous/CMakeLists.txt
@@ -6,4 +6,13 @@ find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 
 project(app LANGUAGES C)
 
+# The Hubble master key is provisioned at build time via the
+# CONFIG_SAMPLE_HUBBLE_KEY Kconfig option (a base64 string), e.g.
+#
+#   west build -b <board> . -- -DCONFIG_SAMPLE_HUBBLE_KEY="<base64 key>"
+#
+# It is decoded into bytes at runtime (Zephyr provides base64_decode). If it is
+# omitted (e.g. a CI build-only run) the firmware falls back to an all-zero
+# placeholder. See src/main.c and Kconfig.
+
 target_sources(app PRIVATE src/main.c)
--- a/samples/zephyr/sat-continuous/README.md
+++ b/samples/zephyr/sat-continuous/README.md
@@ -1,0 +1,107 @@
+# Hubble Network Satellite Continuous Sample
+
+This sample continuously builds Hubble satellite packets and transmits them in a
+loop. Use it as a starting point for satellite (space) connectivity on a
+supported radio.
+
+The sample uses the **device uptime** counter source
+(`CONFIG_HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME`), so it does not need UTC time
+provisioned: the EID counter used for encryption starts at 0 and advances with
+device uptime. Only the master key has to be embedded before building.
+
+> [!NOTE]
+> Satellite functionality is pre-production and not yet ready for production
+> deployments.
+
+## Requirements
+
+- A cryptographic key provided by Hubble Network.
+- A supported satellite-capable board (see [Supported boards](#supported-boards)).
+- For Nordic and Silicon Labs targets: the prebuilt radio blob (see
+  [Radio blobs](#radio-blobs)).
+
+## Radio blobs
+
+The satellite radio implementation for Nordic SoCs ships as a prebuilt static
+library (a "blob") rather than source. These are **not** fetched by
+`west update`; you must pull them in explicitly once after setting up the
+workspace:
+
+```sh
+west blobs fetch hubblenetwork-sdk
+```
+
+This downloads the satellite libraries for the nRF52/nRF53/nRF54 families into
+the module's `zephyr/blobs/` directory. If the blob for your target is missing,
+the build fails to link the satellite radio. Run `west blobs list
+hubblenetwork-sdk` to see what is available and whether it has been fetched.
+
+Silicon Labs targets (`xg24_rb4187c`, `xiao_mg24`) use the in-tree Gecko custom
+radio PHY (`CONFIG_SOC_GECKO_CUSTOM_RADIO_PHY`), but that PHY links against
+Silicon Labs' RAIL library, which is itself distributed as a blob. Fetch it once
+before building for those boards:
+
+```sh
+west blobs fetch hal_silabs
+```
+
+## Building and running
+
+Pass your base64 master key (the string Hubble gave you, already base64) on the
+build command line with `-DCONFIG_SAMPLE_HUBBLE_KEY`. The string is embedded in the firmware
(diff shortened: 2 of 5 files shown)


samples/zephyr/sat-continuous/CMakeLists.txt
```diff
@@ -6,4 +6,39 @@ find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
...
+      "import base64,sys; print(','.join(hex(b) for b in base64.b64decode(sys.argv[1])))"
+      "${HUBBLE_KEY}"
+    OUTPUT_VARIABLE KEY_BYTES
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+    RESULT_VARIABLE decode_result)
+  if(NOT decode_result EQUAL 0)
+    message(FATAL_ERROR
+      "Failed to base64-decode HUBBLE_KEY; check that the key is valid base64.")
+  endif()
+endif()
```
ceolin: I don't like this very much, this is a Zephyr sample and Zephyr has base64 available. We could just do something like:
```
	if (strlen(HUBBLE_KEY) != 0) {
		size_t olen;

		err = base64_decode(_hubble_key, sizeof(_hubble_key), &olen,
				    HUBBLE_KEY,
				    strlen(HUBBLE_KEY));
		if (err != 0) {
			LOG_ERR("Invalid key provided!");
			return -EINVAL;
		}
	}
```

samples/zephyr/sat-continuous/prj.conf
```diff
@@ -3,4 +3,15 @@
...
+# Use the device uptime counter source so the device does not need UTC time
+# provisioned. The EID counter starts from the value passed to hubble_init()
+# and advances with uptime.
+CONFIG_HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME=y
+
+# Disable nonce-reuse enforcement for this sample. The check persists the
+# sequence counter to NVS and rejects reused/decreasing counters; this sample
+# transmits continuously and needs no NVS, so turn it off.
+CONFIG_HUBBLE_NETWORK_SECURITY_ENFORCE_NONCE_CHECK=n
+
```
ceolin, line 22: NVS is just across reboots, why not just not reduce the frequency new packets are created ?

Just saying because it is convenient to just disable it instead of doing the right thing and we don't want to incentivize it.

ceolin (dismissed):
Nothing to block, but please consider the comments.

samples/zephyr/sat-continuous/src/main.c
```diff
@@ -65,13 +74,13 @@ int main(void)
 			goto end;
 		}
 
-		err = hubble_sat_packet_send(&pkt, HUBBLE_SAT_RELIABILITY_NORMAL);
+		err = hubble_sat_packet_send(&pkt, HUBBLE_SAT_RELIABILITY_NONE);
```
HongNguyen635, line 117: what I'm a bit concerned is people is likely to copy the sample code, and copy this part. Although we have the reliability NONE, we often don't want people to use this and instead, use the recommended profile.

samples/zephyr/sat-continuous/CMakeLists.txt
HongNguyen635: I agree with this, we can just add a Kconfig and let user pass in the key and decode it.

samples/zephyr/sat-continuous/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+This downloads the satellite libraries for the nRF52/nRF53/nRF54 families into
+the module's `zephyr/blobs/` directory. If the blob for your target is missing,
+the build fails to link the satellite radio. Run `west blobs list
+hubblenetwork-sdk` to see what is available and whether it has been fetched.
+
+> [!NOTE]
+> Silicon Labs targets (`xg24_rb4187c`, `xiao_mg24`) use the in-tree Gecko
+> custom radio PHY (`CONFIG_SOC_GECKO_CUSTOM_RADIO_PHY`) and do not require a
+> blob fetch.
```
HongNguyen635: they still requires `west blobs fetch hal_silabs` because RAIL is provided as blob. But I guess we are not required to mention it here (?)

samples/zephyr/sat-continuous/src/main.c
buckleypaul: This is more accurate what we want for this demo though - this is explicitly an app that we want to continually transmit at some X rate. Current logic actually isn't very correct and is somewhat fragile to us changing the logic for high and default reliability (which I strongly believe we _will_ want to adjust and tune over time). This application shouldn't be coupled to TDR or transmission mechanism logic.

samples/zephyr/sat-continuous/CMakeLists.txt
buckleypaul: I'm mixed - given that we can error on user input, it's sort of nice having it error at this point vs in the code where errors can be silent. Not strongly opinionated though - I've seen people do things like remove the = or only past part of the key and it's nice having the error at build time.

samples/zephyr/sat-continuous/prj.conf
buckleypaul: It's nice having a one-to-one for packets transmitted sometimes - same reason we're doing it in the soak test.

samples/zephyr/sat-continuous/src/main.c
HongNguyen635: That makes sense, I'm not really opinionated about this. Just something I'd make a mistake in doing if this is my first time working with Hubble.
I guess we can put a comment directly above this line that said using NORMAL / HIGH profile is recommended, NONE is just for testing purpose would suffice.

samples/zephyr/sat-continuous/README.md
buckleypaul: Updated the readme

samples/zephyr/sat-continuous/CMakeLists.txt
buckleypaul: I made it at runtime but still have it something passed when building vs a kconfig. Could have it do either if we wanted, but I personally prefer not needing to edit a file and it makes it scriptable which is sorta nice.

samples/zephyr/sat-continuous/src/main.c
buckleypaul: Done

HongNguyen635 (approved):

PR #245, opened 2026-06-15
sat: zephyr: nrf: Refactor DTM code

ceolin:
Move the whole DTM code to a common file to avoid duplications.


port/zephyr/boards/nordic/nrf54/nrf54_soc.c
HongNguyen635: remove the `static` as well?

port/zephyr/boards/nordic/sat_dtm.c
```diff
@@ -0,0 +1,212 @@
...
+#define RADIO_TXPOWER_TXPOWER_Pos2dBm (0x02UL)
+#endif /* RADIO_TXPOWER_TXPOWER_Pos2dBm */
+
+#ifndef RADIO_TXPOWER_TXPOWER_Pos1dBm
+#define RADIO_TXPOWER_TXPOWER_Pos1dBm (0x01UL)
+#endif /* RADIO_TXPOWER_TXPOWER_Pos1dBm */
+
+#endif /* CONFIG_SOC_COMPATIBLE_NRF5340_CPUNET */
+
+extern nrf_radio_txpower_t _sat_power;
```
HongNguyen635, line 38: do we want to rename this to `sat_power` now since it's no longer static?

HongNguyen635 (commented):
Nice, confirm that nRF21 is still working after the change 👍

port/zephyr/boards/nordic/sat_dtm.c
ceolin: I think this symbol should continue to be private, not glad with these variable access, but it is ok for now imho.

HongNguyen635 (approved):

PR #249, opened 2026-06-17
tools: add orbital_params_fetch.py

ceolin:
Add a script that fetches orbital params from the Hubble satellite API and generates a C struct file (sat_params.c) for the satellites.


tools/orbital_params_fetch.py
```diff
@@ -0,0 +1,101 @@
...
+where <prefix> is an optional directory prefix (default: current directory).
+"""
+
+import argparse
+import os
+from pathlib import Path
+
+import requests
+
+API_BASE = "https://api.hubble.com/api/satellite"
```
HongNguyen635, line 18: This returns a 404, did you mean`https://api.hubble.com/api/satellite/ephemeris`?

tools/orbital_params_fetch.py
```diff
@@ -0,0 +1,101 @@
...
+TOKEN = os.environ["HUBBLE_API_TOKEN"]
+
+TARGET_SAT_IDS = {59146, 60471, 64562, 64565, 64592, 64840}
```
HongNguyen635: I'm thinking we should remove this list and just parse whatever the backend returns to us. Currently the backend is limited to certain number of sats, so depending on which org & entitlement, they will have different sats return to them.

tools/orbital_params_fetch.py
```diff
@@ -0,0 +1,101 @@
+"""
+Fetch orbital params from Hubble API.
+
+Writes a C struct file for the orbital params to <prefix>/sat_params.c,
+where <prefix> is an optional directory prefix (default: current directory).
```
HongNguyen635, line 5: we can probably add a `usage` line here so that user knows what to call (e.g. set env variable and the arguments).

tools/orbital_params_fetch.py
```diff
@@ -0,0 +1,101 @@
...
+def main():
+    args = parse_args()
+    output_path = args.prefix / SAT_PARAMS_RELPATH
+
+    ephemeris = fetch_json("ephemeris")
+    orbital_params = list(iter_target_orbital_params(ephemeris["satellites"]))
+
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+    output_path.write_text(render_sat_params(orbital_params))
+    print(f"Wrote {output_path}")
```
HongNguyen635: Maybe log the number of sat orb params written as well?

buckleypaul:
Are you planning to add testing for this? Can easily get out of sync (not necessary for today).

tools/orbital_params_fetch.py
ceolin: the endpoint is appended to this.

HongNguyen635 (approved):

PR #250, opened 2026-06-17
api: use milliseconds for satellite pass prediction times

buckleypaul:
The rest of the SDK exchanges time in milliseconds (uptime, time set, transmission period), so accepting and returning seconds here forced callers to special-case unit conversion around these APIs. Align the pass prediction API with that convention so all SDK time values share one unit.

This is a breaking change for integrators currently passing or reading seconds.

### DIFF
--- a/samples/freertos/ti/sat-dual-stack/src/main.c
+++ b/samples/freertos/ti/sat-dual-stack/src/main.c
@@ -35,6 +35,7 @@ icall_userCfg_t user0Cfg = BLE_USER_CFG;
 #include "app_ble.h"
 
 #define US_PER_SEC 1000000ULL
+#define US_PER_MS  1000ULL
 
 #pragma message(                                                               \
 	"Dummy key, replace with actual key by running ./embed_key_time.py -b b64key -d")
@@ -90,7 +91,7 @@ void *main_thread_entry(void *arg0)
 
 	struct hubble_sat_pass_info pass_info;
 	struct hubble_sat_packet packet;
-	uint64_t now_s;
+	uint64_t now_ms;
 	uint64_t sat_wait_us;
 	bStatus_t status;
 	int ret;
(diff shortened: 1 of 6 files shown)


ceolin:
missing change https://github.com/HubbleNetwork/hubble-device-sdk/blob/5a42cfb397c7c921e51f51e2c4e50c3dde4e990e/samples/zephyr/sat-dual-stack/src/main.c#L140

samples/freertos/ti/sat-dual-stack/src/main.c
```diff
@@ -154,10 +155,11 @@ void *main_thread_entry(void *arg0)
 		Log_printf(
 			Log_Dual_Stack,
 			Log_INFO, "Next pass at: %u (unix epoch seconds), max elevation angle: %.2f",
-			pass_info.start, pass_info.max_elevation_angle);
+			(uint32_t)(pass_info.start / 1000U),
```
HongNguyen635, line 158: you don't need to do this right? Just change the log line unit to ms.

samples/zephyr/sat-dual-stack/src/main.c
```diff
@@ -151,8 +151,8 @@ int main(void)
 		k_timer_start(&sat_timer, K_SECONDS(120), K_NO_WAIT);
 #else
 		LOG_INF("Next pass at %llu (unix epoch seconds)",
-			pass_info.start);
-		k_timer_start(&sat_timer, K_SECONDS(pass_info.start - now_s),
+			pass_info.start / MSEC_PER_SEC);
```
HongNguyen635, line 154: same thing here

HongNguyen635 (commented):
some minor comments but I'll need to re rerun the tests on the hw to make sure nothing is breaking.

samples/freertos/ti/sat-dual-stack/src/main.c
buckleypaul: It looks like the TI log code can't print a uint64, so either need to truncate or do something different here. IMO it would make more sense to have this print like "X seconds until next pass" vs "next pass at"

HongNguyen635: yeah we can do "X seconds until next pass" if you see that works better. For me I tend to just check clock time and know that at X hour Y min Z sec there is a pass. You could keep the "next pass at" and add the "X seconds until next pass" log as well just to make sure the timer is scheduled correctly.

HongNguyen635 (approved):
I ran a couple of tests again:
- Pass prediction vs pass times from backend.
- Dual stack on nrf54 (for zephyr) and cc27.

They're all look good so I'm not going to block this PR. But please consider the comments above.

samples/freertos/ti/sat-dual-stack/src/main.c
buckleypaul: I'll just keep this as is then - it's best to keep the TI and Zephyr versions the same and since uint64 prints aren't supported I'll just keep this division.

PR #252, opened 2026-06-17
Iniital Sat Continuous HIL Test

hunterhubble:
- Added .yml file to trigger workflow in HITL-endpoint via repository dispatch
- Coincides with a PR in Hitl-Endpoint, which currently only tests zephyr boards

### DIFF
--- a/.github/workflows/hil-sat-continuous.yml
+++ b/.github/workflows/hil-sat-continuous.yml
@@ -1,0 +1,17 @@
+name: HITL Trigger - SAT Continuous
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  trigger-hitl-endpoint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger HITL sat-continuous build/flash
+        uses: peter-evans/repository-dispatch@v3
+        with:
+          token: ${{ secrets.HITL_ENDPOINT_REPO }}
+          repository: HubbleNetwork/hitl-endpoint
+          event-type: sdk-sat-continuous-pr
+          client-payload: '{"ref": "${{ github.event.pull_request.head.sha || github.sha }}", "branch": "${{ github.head_ref || github.ref_name }}"}'


hunterhubble:
A couple of things should happen to make this work. 

1) This PR from hitl-endpoint, which has all of the tests https://github.com/HubbleNetwork/hitl-endpoint/pull/114
2) GH secret to enable the repository dispatch to enable the hitl-endpoint tests

.github/workflows/hil-sat-continuous.yml
```diff
@@ -0,0 +1,17 @@
+name: HITL Trigger - SAT Continuous
```
HongNguyen635, line 1: This 1st step is ok. We could later improve on this workflow to run tests to all samples instead of just sat continuous.

.github/workflows/hil-sat-continuous.yml
```diff
@@ -0,0 +1,17 @@
...
+
+on:
+  pull_request:
+  workflow_dispatch:
+
```
HongNguyen635, line 6: we can be explicit with `permissions contents: read`

.github/workflows/hil-sat-continuous.yml
```diff
@@ -0,0 +1,17 @@
...
+  trigger-hitl-endpoint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger HITL sat-continuous build/flash
+        uses: peter-evans/repository-dispatch@v3
+        with:
+          token: ${{ secrets.HITL_ENDPOINT_REPO }}
+          repository: HubbleNetwork/hitl-endpoint
+          event-type: sdk-sat-continuous-pr
+          client-payload: '{"ref": "${{ github.event.pull_request.head.sha || github.sha }}", "branch": "${{ github.head_ref || github.ref_name }}"}'
```
HongNguyen635, line 17: think about the security comment I put in the HITL repo. Since this is a public repo, doing this could be quite dangerous.

PR #253, opened 2026-06-17
samples: add sat-pass-prediction Zephyr sample

buckleypaul:
A sample application that uses satellite pass prediction to time transmissions. The main loop computes the next pass above the device location, sleeps until it opens, then broadcasts a packet.

An optional shell interface exposes runtime state, and a companion tool fetches orbital parameters from the Hubble ephemeris API.


samples/zephyr/sat-pass-prediction/src/main.c
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	for (;;) {
+		struct hubble_sat_pass_info pass;
+		struct hubble_sat_packet pkt;
+		/* hubble_time_get() is milliseconds; pass prediction wants seconds. */
+		uint64_t now = hubble_time_get() / 1000ULL;
+		int64_t wait;
+
+		err = hubble_sat_next_pass_get(now, &device_pos, &pass);
```
HongNguyen635, line 267: Since we [put in the doc](https://github.com/HubbleNetwork/hubble-device-sdk/blob/15ca87b0b3d14f99d165e7cbd636113e700030b2/docs/satellite/index.rst?plain=1#L103) that device uptime must never be used for pass prediction, I'm not sure if having this sample is a good idea since we're going against our recommendation.

The issue is that `hubble_sat_next_pass_get` account for retries for TDR, with device uptime as counter source, the call to `hubble_internal_time_drift_get()` will returns 0 anyway since time is not synced. So time drift is not accounted for regardless of whatever value `CONFIG_HUBBLE_SAT_NETWORK_DEVICE_TDR` is set to.

HongNguyen635 (commented):
we can get this sample in if this benefits the bring-up flow. If so, can you please update the sample? There has been a couple of changes since then: we no longer need board conf files to select each symbols, make the mock board default to n and select it in the CI build sample.yaml, update pass prediction to ms, etc.

PR #255, opened 2026-06-18
zephyr: blobs: Select satellite library by float ABI

ceolin:
Nordic satellite blobs are now provided for both the hard- and soft-float ABIs. Pick the matching variant at build time based on CONFIG_FPU and load it from the new nrf5x/{hard,soft}/ blob layout.


HongNguyen635 (approved):
Tested and it worked!

PR #258, opened 2026-06-18
doc: satellite network documentation

ceolin:
Adds Satellite Network guide

- Document how the Satellite Network works at a high level.
- Document pass prediction, including orbital parameter provisioning.
- Add common use cases for continuous transmission and dual-stack BLE plus satellite pass prediction.
- Document reliability modes, retry timing, clock drift compensation, and blocking radio behavior.
- ...

### DIFF
--- a/docs/satellite/index.rst
+++ b/docs/satellite/index.rst
@@ -9,6 +9,437 @@ Satellite Network Overview
    ready for production deployments. APIs and behavior may change in future
    releases.
 
+Introduction
+************
+
+The Hubble Satellite Network lets devices send messages through Hubble's
+satellite network without adding a cellular modem or gateway. It is designed
+for devices that need connectivity beyond the terrestrial coverage.
+
+The application is responsible for deciding *when* to transmit. Two common
+patterns are supported:
+
+* **Continuous transmission** for devices that are constantly powered or are in a
+  test environment.
+* **Pass-predicted transmission** for devices that should transmit only when a
+  satellite pass is expected.
+
+Satellite packets can carry payloads of ``0``, ``4``, ``9``, or ``13`` bytes.
+The SDK encrypts the payload using the configured Hubble device key and derives
+the packet identifiers from the SDK time counter.
+
+How It Works
+************
+
+At a high level, a satellite transmission follows this sequence:
+
+1. Initialize the SDK with :c:func:`hubble_init`.
+2. Build a satellite packet with :c:func:`hubble_sat_packet_get`.
+3. Send the packet with :c:func:`hubble_sat_packet_send`.
+4. The platform port enables the radio, converts the packet into frames, sends
+   each frame, and disables the radio when transmission is complete.
+
+The SDK uses the configured counter source for key derivation and packet
+identity:
+
+* ``CONFIG_HUBBLE_COUNTER_SOURCE_UNIX_TIME`` uses Unix time in milliseconds.
+* ``CONFIG_HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME`` uses a device-uptime-derived
+  counter and does not require UTC time.
+
+Use Unix time when the device can provision or synchronize time and needs pass
+prediction. Use device uptime for simple continuous transmission workflows that
+do not need wall-clock scheduling.
+
+Packet Generation
+=================
+
+:c:func:`hubble_sat_packet_get` creates the encrypted satellite packet. The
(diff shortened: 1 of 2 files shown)


docs/satellite/index.rst
```diff
@@ -9,6 +9,427 @@ Satellite Network Overview
...
+The SDK uses the configured counter source for key derivation and packet
+identity:
+
+* ``CONFIG_HUBBLE_COUNTER_SOURCE_UNIX_TIME`` uses Unix time in milliseconds.
+* ``CONFIG_HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME`` uses a device-uptime-derived
+  counter and does not require UTC time.
+
+Use Unix time when the device can provision or synchronize time and needs pass
+prediction. Use device uptime for simple continuous transmission workflows that
+do not need wall-clock scheduling.
```
HongNguyen635, line 51: The use case of device uptime + pass prediction might not happen at all, but should we add a warning that said, when using device uptime as counter source, pass prediction must never be used because the retries compensation takes time sync into account?

docs/satellite/index.rst
```diff
@@ -9,6 +9,427 @@ Satellite Network Overview
...
+
+:c:func:`hubble_sat_packet_send` sends a prepared packet with the selected
+reliability mode:
+
+.. code-block:: c
+
+   err = hubble_sat_packet_send(&packet, HUBBLE_SAT_RELIABILITY_NORMAL);
+
+The function is blocking. It returns after the full transmission sequence has
+completed or after an error is reported by the platform radio port.
```
HongNguyen635, line 88: This is mentioned later in more detail, but we can also put it briefly here that the function also takes retries into account for normal and high reliability mode.

docs/satellite/index.rst
```diff
@@ -9,6 +9,427 @@ Satellite Network Overview
...
+   memcpy(&sat->eccentricity, p + 64, sizeof(double));
+   memcpy(&sat->satellite_id, p + 72, sizeof(uint32_t));
+
+When receiving binary provisioning data, copy each field separately as shown
+above. This avoids unaligned accesses on MCUs that do not support them.
+
+For devices that do not receive orbital parameters at runtime, you can generate
+a C source file with hard-coded satellite parameters and build it into the
+application. The ``tools/orbital_params_fetch.py`` helper fetches current
+orbital parameters from the Hubble API and writes ``sat_params.c``.
```
HongNguyen635, line 160: Need to add a pro/con of this method. e.g. how long is the data valid for. I imagine people will ask this question a lot.

docs/satellite/index.rst
```diff
@@ -9,6 +9,427 @@ Satellite Network Overview
...
+
+.. code-block:: c
+
+   if (pass.start <= now_s) {
+           err = hubble_sat_next_pass_get(pass.start + pass.duration,
+                                          &pos, &pass);
+   }
+
+For region-based scheduling, use :c:func:`hubble_sat_next_pass_region_get` with
+``struct hubble_sat_device_region``.
```
HongNguyen635: we probably need to add a note that this func is currently experimental and not encourage for real usage.

docs/satellite/index.rst
```diff
@@ -9,6 +9,427 @@ Satellite Network Overview
...
+
+1. Start BLE provisioning.
+2. Receive Unix time, device location, and satellite orbital parameters.
+3. Initialize the Hubble SDK with the provisioned Unix time and key.
+4. Register orbital parameters with :c:func:`hubble_sat_satellites_set`.
+5. Compute the next pass with :c:func:`hubble_sat_next_pass_get`.
+6. Start BLE beaconing while waiting for ``pass.start``.
+7. Stop BLE when the pass timer expires.
+8. Build and send a satellite packet.
+9. Repeat from pass calculation.
```
HongNguyen635: Maybe `9. Repeat from pass calculation (step 5).`

HongNguyen635 (commented):
Just some minor comments, but the overal lgtm.

docs/satellite/index.rst
ceolin: It is probably better to remove it, I added it for completeness sake but at this point I've changed my mind. It is probably better not having it

docs/satellite/index.rst
ceolin: I don't want to enter this realm now because it may depend on product directions. We can revisit it later imho

HongNguyen635 (dismissed):

HongNguyen635 (approved):

PR #261, opened 2026-06-23
samples: ti: sat: Add option to provision key and time

hunterhubble:
- Add an option to use embed_key_time in the way already described in sat-dual-stack to provision key and time. for sat-dual-stack and sat-continuous
- Helpful for testing so we can detect packets with the correct device ID

### DIFF
--- a/samples/freertos/ti/sat-continuous/README.md
+++ b/samples/freertos/ti/sat-continuous/README.md
@@ -46,8 +46,17 @@ export TICLANG_ARMCOMPILER=/path/to/ti/ti-cgt-armllvm
 export SIMPLELINK_LOWPOWER_F3_SDK_INSTALL_DIR=/path/to/ti/simplelink_lowpower_f3_sdk
 export SYSCONFIG_TOOL=/path/to/ti/sysconfig/sysconfig_cli.sh
 ```
+2. **Provision Time and Key (Optional)**
 
-2. **Build the Project**
+The device's Hubble key must be baked into the firmware at build time. Use the
+*embed_key_time.py* script to generate the key in hex and time in Unix. The tool will generate
+`key.c` and `time.c` into your project's /src directory.
+
+```bash
+python ../../../../tools/embed_key_time.py --base64 <path-to-key> -o src
+```
+
+3. **Build the Project**
 
    Build the project using the provided *makefile*:
 
@@ -57,7 +66,7 @@ make -f cc2340r5.mk
 # or make -f cc2755p10.mk
 ```
 
-3. **Flash the Firmware**
+4. **Flash the Firmware**
 
    Flash the generated firmware (*sat-continuous.out*) onto the target device using your preferred flashing tool.
 
--- a/samples/freertos/ti/sat-continuous/src/main.c
+++ b/samples/freertos/ti/sat-continuous/src/main.c
@@ -23,15 +23,25 @@
 
 #define SLEEP_PERIOD_MS 1000
 
-static uint64_t _unix_time = 0xdeadbeef;
-static uint8_t _key_unused[CONFIG_HUBBLE_KEY_SIZE];
+#if defined(HUBBLE_KEY_TIME_SET)
+#include "key.c"
+#include "time.c"
+#else
+
+#pragma message(                                                               \
+	"Dummy key, replace with actual key by running ./embed_key_time.py -b b64key to set key and time")
+
+static uint8_t master_key[CONFIG_HUBBLE_KEY_SIZE];
+static uint64_t unix_time = 0xdeadbeef;
+
+#endif
 
 void *mainThread(void *arg0)
 {
 	struct hubble_sat_packet packet;
 	int ret;
 
-	ret = hubble_init(_unix_time, _key_unused);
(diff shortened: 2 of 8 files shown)


samples/freertos/ti/sat-continuous/src/main.c
```diff
@@ -23,15 +23,26 @@
 
 #define SLEEP_PERIOD_MS 1000
 
-static uint64_t _unix_time = 0xdeadbeef;
-static uint8_t _key_unused[CONFIG_HUBBLE_KEY_SIZE];
+#if __has_include("key.c")
+#include "key.c"
+#else
+#pragma message(                                                               \
+	"Dummy key, replace with actual key by running ./embed_key_time.py -b b64key -d")
```
HongNguyen635: You can just check for both key and time here, since the `embed_key_time.py` generate both. Also removes the `-d` flag if we go this way.

HongNguyen635 (commented):
kind of a conincident since we just talked about it in [PR201](https://github.com/hubbleNetwork/hubble-device-sdk/pull/201#discussion_r3311734272).

using `__has_include` is ok, just note that it is compiler dependent. If you want consistent you can do similar to what the `ble_beacon` does fd4711dffd6978fdbdd9ce6a04703a4a8a559af4

samples/freertos/ti/sat-dual-stack/src/main.c
```diff
@@ -37,9 +37,19 @@ icall_userCfg_t user0Cfg = BLE_USER_CFG;
...
+#include "key.c"
+#else
 #pragma message(                                                               \
 	"Dummy key, replace with actual key by running ./embed_key_time.py -b b64key -d")
-static uint8_t _hubble_key[CONFIG_HUBBLE_KEY_SIZE];
+static uint8_t master_key[CONFIG_HUBBLE_KEY_SIZE];
+#endif
+
+#if __has_include("time.c")
+#include "time.c"
```
HongNguyen635: using time.c here is unecessary because the sample waits for time, location, and orbital params.

samples/freertos/ti/sat-dual-stack/src/main.c
HongNguyen635: this won't work as expected, because the 1st BLE advertisement is not a hubble non-connectable packet, but it is connectable advertisement that is meant to used with `dual-stack-companion.py`
And time will be reset to 0 later here because of the workaround.

samples/freertos/ti/sat-continuous/src/main.c
hunterhubble: Good point, fixed

samples/freertos/ti/sat-dual-stack/src/main.c
hunterhubble: Gotcha, fixed

samples/freertos/ti/sat-dual-stack/src/main.c
hunterhubble: I understand, the sample will only check for the key now

HongNguyen635 (dismissed):

HongNguyen635 (commented):
Can you please update the README as well?

1. Add instruction for embedded key and time in `sat-cont` [README](https://github.com/HubbleNetwork/hubble-device-sdk/blob/main/samples/freertos/ti/sat-continuous/README.md)
2. Update this step in `sat-dual-stack` [README](https://github.com/HubbleNetwork/hubble-device-sdk/blob/15ca87b0b3d14f99d165e7cbd636113e700030b2/samples/freertos/ti/sat-dual-stack/README.md?plain=1#L58) to match the variable name and python script argument.

samples/freertos/ti/sat-continuous/README.md
```diff
@@ -46,8 +46,16 @@ export TICLANG_ARMCOMPILER=/path/to/ti/ti-cgt-armllvm
...
+2. **Provision Time and Key (Optional)**
 
-2. **Build the Project**
+The device's Hubble key must be baked into the firmware at build time. Use the
+*embed_key_time.py* script to generate the key in hex and time in Unix. The tool will generate
+key.c into your project's /src directory.
+
+```bash
+python ../../../../tools/embed_key_time.py --base64 <path-to-key> -o src
+
```
HongNguyen635: missing the closing ```
also nit: `src/` and ` ` around key.c

samples/freertos/ti/sat-dual-stack/README.md
```diff
@@ -58,11 +58,11 @@ pip install -r ../../../../tools/requirements-companion.txt
...
 
 The device's Hubble key must be baked into the firmware at build time. Use the
-*embed_key_time.py* script to generate the key in hex, and copy that into
-`_hubble_key` array:
+*embed_key_time.py* script to generate the key in hex. The tool will generate
+key.c into your project's /src directory.
 
 ```bash
-python ../../../../tools/embed_key_time.py --base64 <path-to-key> -d
+python ../../../../tools/embed_key_time.py --base64 <path-to-key> -o src
```
HongNguyen635: nit: `src/` and \` \` around key.c

HongNguyen635 (approved):

PR #264, opened 2026-06-26
zephyr: Kconfig: Simplify satellite symbols

ceolin:
Centralizes the SoC-specific Kconfig defaults required for the Hubble SAT +
BLE dual-stack and replaces the per-board SoftDevice overlay with a reusable Zephyr snippet. Previously these defaults were copy-pasted into every sample's board `.conf` file; this PR moves them into the SDK's Kconfig so samples stay minimal and the configuration stays consistent across targets.


docs/_extensions/kconfig_autodoc.py
```diff
@@ -22,55 +22,84 @@
 import logging
 import os
 import re
+import sys
 import tempfile
```
HongNguyen635: we can remove the `re` and `tempfile` import since these are no longer used.

samples/zephyr/sat-continuous/boards/nrf21540dk.conf
```diff
@@ -1,7 +1 @@
-CONFIG_NRFX_TIMER=y
-CONFIG_SAMPLE_PROVIDE_SAT_BOARD_SUPPORT=n
-CONFIG_BT_CTLR_ADVANCED_FEATURES=y
-CONFIG_BT_CTLR_DYNAMIC_INTERRUPTS=y
-CONFIG_DYNAMIC_INTERRUPTS=y
-CONFIG_DYNAMIC_DIRECT_INTERRUPTS=y
-CONFIG_SHARED_INTERRUPTS=y
+CONFIG_SAMPLE_PROVIDE_SAT_BOARD_SUPPORT=n
```
HongNguyen635, line 1: should we make this symbol `n` by default and y-selected it in the test case? It feels kind of odd to have a lot of conf files with only this option.

HongNguyen635 (commented):
A few minor comments.

samples/zephyr/sat-continuous/boards/nrf21540dk.conf
ceolin: I can revisit it, the main issue was CI integration, but I will review it again.

docs/_extensions/kconfig_autodoc.py
ceolin: good catch.

samples/zephyr/sat-continuous/boards/nrf21540dk.conf
HongNguyen635: as in a different PR or this same one?

ceolin: different pr, that is a different thing to address

HongNguyen635 (approved):

PR #266, opened 2026-06-30
samples: zephyr: default mock satellite board support to off

ceolin:
Make CONFIG_SAMPLE_PROVIDE_SAT_BOARD_SUPPORT default to n in both the sat-continuous and sat-dual-stack samples, so they build with not mock board APIs. Twister test cases that exercise satellite APIs now opt in explicitly via extra_configs.


buckleypaul (approved):

samples/zephyr/sat-dual-stack/Kconfig
```diff
@@ -19,7 +19,6 @@ config HUBBLE_SAMPLE_DEBUG
 
 config SAMPLE_PROVIDE_SAT_BOARD_SUPPORT
     bool "Provide satellite board suport for test purpose"
-    default y
```
HongNguyen635, line 22: should we just do `default n` instead of removing it? The `n` is implicit but the other config states the default option regardless of `y` or `n` (for example the one above this).

HongNguyen635 (commented):
Still need some minor edits to the [sat-continuous README](https://github.com/HubbleNetwork/hubble-device-sdk/blob/cce770eb2a8f82c8d2fea2bcaee9800f190868d1/samples/zephyr/sat-continuous/README.md?plain=1#L106) and [sat-dual-stack README](https://github.com/HubbleNetwork/hubble-device-sdk/blob/cce770eb2a8f82c8d2fea2bcaee9800f190868d1/samples/zephyr/sat-dual-stack/README.md?plain=1#L32) to mention that the default is now `n`.

samples/zephyr/sat-dual-stack/Kconfig
ceolin: Zephyr usually don't set `default n` and I followed it here. I noticed the other but was not related with these changes.

PR #270, opened 2026-07-06
sat: Add hubble_sat_packet_send_pass

ceolin:
Add `hubble_sat_packet_send_pass()` API that accepts a `hubble_sat_pass_info` to drive the transmission window instead of relying on the fixed retry counts defined by the transmission mode

### DIFF
--- a/docs/satellite/reliability.rst
+++ b/docs/satellite/reliability.rst
@@ -3,10 +3,11 @@
 Reliability and Power Consumption
 #################################
 
-The reliability mode passed to :c:func:`hubble_sat_packet_send` controls how
-many times the SDK asks the platform radio port to send the same packet and how
-far apart those transmissions are spaced. It is the primary trade-off between
-delivery probability and energy use.
+The reliability mode passed to :c:func:`hubble_sat_packet_send` and
+:c:func:`hubble_sat_packet_pass_send` controls how many times the SDK asks the
+platform radio port to send the same packet and how far apart those
+transmissions are spaced. It is the primary trade-off between delivery
+probability and energy use.
 
 Reliability Modes
 *****************
@@ -59,3 +60,46 @@ For battery-powered products, the most power-efficient design is usually a
 pass-predicted workflow: sleep or use low-power BLE between passes, wake before
 ``pass.start``, transmit with the lowest reliability mode that meets the product
 delivery requirement, then return to the low-power state.
+
+Pass-Adaptive Transmission
+***************************
+
+:c:func:`hubble_sat_packet_pass_send` sends a packet using a
+:c:struct:`hubble_sat_pass_info` window instead of the fixed baselines in the
+table above. Rather than always transmitting the mode's baseline number of
+times, it derives the retry count from ``pass.duration`` so the packet is
+retransmitted as many times as will fit while the satellite is overhead.
+
+.. code-block:: c
+
+   struct hubble_sat_pass_info pass;
+
+   err = hubble_sat_next_pass_region_get(hubble_time_get(), &region, &pass);
+   if (err != 0) {
+           return err;
+   }
+
+   err = hubble_sat_packet_pass_send(&packet, HUBBLE_SAT_RELIABILITY_NORMAL,
+                                     &pass);
+   if (err != 0) {
+           return err;
+   }
+
+* **Required for a region pass.** A pass obtained from
+  :c:func:`hubble_sat_next_pass_region_get` has no single point of reference
+  the fixed baselines were tuned for, so :c:func:`hubble_sat_packet_send` has
+  no way to fit its retries to that window. Use
+  :c:func:`hubble_sat_packet_pass_send` whenever the pass came from a region
+  query.
+* **Also usable for a single-point pass.** A pass from
+  :c:func:`hubble_sat_next_pass_get` works too — in that case
+  :c:func:`hubble_sat_packet_pass_send` is a drop-in alternative to
+  :c:func:`hubble_sat_packet_send` that adapts retries to that specific pass's
+  duration instead of using the mode's fixed baseline.
+* **``HUBBLE_SAT_RELIABILITY_NONE`` is invalid.** This function requires a
+  mode that retries (``HUBBLE_SAT_RELIABILITY_NORMAL`` or
+  ``HUBBLE_SAT_RELIABILITY_HIGH``); passing ``HUBBLE_SAT_RELIABILITY_NONE``, a
+  NULL ``packet``, or a NULL ``pass`` returns ``-EINVAL``.
+
+See :ref:`hubble_pass_prediction_best_practices` for how to obtain a
+:c:struct:`hubble_sat_pass_info`.
--- a/include/hubble/sat.h
+++ b/include/hubble/sat.h
@@ -11,6 +11,7 @@
 #include <stdint.h>
 
 #include <hubble/sat/packet.h>
+#include <hubble/sat/pass_prediction.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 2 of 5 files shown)


src/hubble_sat.c
```diff
@@ -150,6 +150,44 @@ int hubble_sat_packet_send(const struct hubble_sat_packet *packet,
...
+	ret = _transmission_params_get(mode, &retries, &interval_s);
+	if (ret < 0) {
+		HUBBLE_LOG_WARNING("Invalid mode given");
+		return ret;
+	}
+
+	if (interval_s > 0U) {
+		retries = (uint8_t)HUBBLE_MIN(
+			UINT8_MAX,
+			HUBBLE_MAX(1U, pass->duration / (1000U * interval_s)));
```
HongNguyen635, line 173: Since `macros.h` is already included, we can use the `HUBBLE_MSEC_PER_SEC` in instead of `1000U`

src/hubble_sat.c
```diff
@@ -150,6 +150,44 @@ int hubble_sat_packet_send(const struct hubble_sat_packet *packet,
...
+		return -EINVAL;
+	}
+
+	ret = _transmission_params_get(mode, &retries, &interval_s);
+	if (ret < 0) {
+		HUBBLE_LOG_WARNING("Invalid mode given");
+		return ret;
+	}
+
+	if (interval_s > 0U) {
```
HongNguyen635, line 170: I'm wondering if it makes sense to reject the reliability none mode since the purpose of this function is to filled up the pass with the retries. None reliability only tx once, and might be kind of misleading.

include/hubble/sat.h
```diff
@@ -87,6 +88,34 @@ enum hubble_sat_transmission_mode {
...
+/**
+ * @brief Transmit a packet during a predicted satellite pass.
+ *
+ * This function sends a packet over the satellite communication channel,
+ * for the duration of time provided in the satellite pass info.
+ * The pass duration is used to derive the number of retransmissions,
+ * so the packet is transmitted as many times as possible while the satellite
+ * is overhead.
+ *
+ * @note This function is blocking: it does not return until the transmission
```
buckleypaul, line 100: It's unclear to me _when_ this function should be called. Is it just called whenever?

include/hubble/sat.h
buckleypaul, line 57: This is actually a false statement given that we adjust for the number with TDR adjustments?

src/hubble_sat.c
```diff
@@ -150,6 +150,44 @@ int hubble_sat_packet_send(const struct hubble_sat_packet *packet,
 	return 0;
 }
 
+int hubble_sat_packet_send_pass(const struct hubble_sat_packet *packet,
```
buckleypaul, line 153: Are we not doing any TDR in this function or is that just already handled via the pass duration?

src/hubble_sat.c
```diff
@@ -150,6 +150,44 @@ int hubble_sat_packet_send(const struct hubble_sat_packet *packet,
...
+	}
+
+	ret = _transmission_params_get(mode, &retries, &interval_s);
+	if (ret < 0) {
+		HUBBLE_LOG_WARNING("Invalid mode given");
+		return ret;
+	}
+
+	if (interval_s > 0U) {
+		retries = (uint8_t)HUBBLE_MIN(
```
buckleypaul, line 171: IMO we should enable the customer to set a maximum number of transmissions and if that is less than what we calculate, we spread the transmissions over that window as best we can. So far both customers, and I could see essentially all customers, want/would want to set a max threshold for power budget calculation reasons. 

They could just control this entirely via the no-reliability mode but this is much simpler for them and seems pretty easy for us to add.

include/hubble/sat.h
ceolin: Not part of this change, but there is a note explaining how TDR impacts / change it.

src/hubble_sat.c
ceolin: Yep, I think I agree with it.

src/hubble_sat.c
ceolin: It may make sense, we can add a new function for this instead of add this logic to current APIs, since that is likely to be an exception and not common use case. Some API to set a max re-transmissions value.

src/hubble_sat.c
ceolin: This is done in the pass prediction computation. We accommodate the drift between start / end of the window time.

include/hubble/sat.h
ceolin: I've added some notes about it. Hope it clarifies the problem.

HongNguyen635 (commented):
Do you think `hubble_sat_packet_send_pass` is a little confusing? Maybe `hubble_sat_packet_send_for_pass` or `hubble_sat_packet_send_over_pass` would make more sense since we're sending a packet "for" a pass.
Not very opinionated though as everything else looks good. I just feel like "send pass" is a bit ambiguous as it sounds like "send pass to... what?"

HongNguyen635 (dismissed):
other than comment above this lgtm.

docs/satellite/reliability.rst
```diff
@@ -59,3 +60,46 @@ For battery-powered products, the most power-efficient design is usually a
...
+  :c:func:`hubble_sat_packet_send_pass` is a drop-in alternative to
+  :c:func:`hubble_sat_packet_send` that adapts retries to that specific pass's
+  duration instead of using the mode's fixed baseline.
+* **``HUBBLE_SAT_RELIABILITY_NONE`` is invalid.** This function requires a
+  mode that retries (``HUBBLE_SAT_RELIABILITY_NORMAL`` or
+  ``HUBBLE_SAT_RELIABILITY_HIGH``); passing ``HUBBLE_SAT_RELIABILITY_NONE``, a
+  NULL ``packet``, or a NULL ``pass`` returns ``-EINVAL``.
+
+See :ref:`hubble_pass_prediction_best_practices` for how to obtain a
+:c:struct:`hubble_sat_pass_info`.
```
HongNguyen635, line 105: maybe put note / warning here to keep the `hubble_sat_pass_info` valid until the call to `hubble_sat_packet_send_pass`?

ceolin: That is true for any parameter you pass through reference :)

HongNguyen635: Yes that is true. But I think this could be common pitfall, but leave as is is ok, that's why I'm not blocking.

ceolin:
> Do you think `hubble_sat_packet_send_pass` is a little confusing? Maybe `hubble_sat_packet_send_for_pass` or `hubble_sat_packet_send_over_pass` would make more sense since we're sending a packet "for" a pass. Not very opinionated though as everything else looks good. I just feel like "send pass" is a bit ambiguous as it sounds like "send pass to... what?"

I don't like it very much too, but these alternatives have the same problem for me that is not ending with the verb as other APIs. What you think about:

`hubble_sat_packet_pass_send()`
`hubble_sat_packet_and_pass_send()`
`hubble_sat_packet_during_pass_send()`
`hubble_sat_packet_over_pass_send()`

I tend to prefer the first between these because is less verbose / shorter, but it is not as clear as the last two.

HongNguyen635:
I agree that `hubble_sat_packet_pass_send` is not as clear as the last 2.
I kind of prefer either `hubble_sat_packet_pass_send` or `hubble_sat_packet_during_pass_send` (or `hubble_sat_packet_per_pass_send`?)

Let's see what others think about which one is better. @buckleypaul @hunterhubble @Micahscope ?

ceolin:
> I agree that `hubble_sat_packet_pass_send` is not as clear as the last 2. I kind of prefer either `hubble_sat_packet_pass_send` or `hubble_sat_packet_during_pass_send` (or `hubble_sat_packet_per_pass_send`?)
> 
> Let's see what others think about which one is better. @buckleypaul @hunterhubble @Micahscope ?

My preference goes to `hubble_sat_packet_pass_send` because follow the namespace logic but   all of them have drawbacks.

hunterhubble:
I agree with hubble_sat_packet_pass_send() if we want to keep with ending with a verb. hubble_sat_packet_send_during_pass() is an option too, it would be a little more descriptive but also wordy and it doesn't follow the convention. Not too opinionated either way as long as the documentation is clear

Micahscope:
Same ideas as Hunter for me. I think `hubble_sat_packet_pass_send()` is best.

Adding `during`, to me, doesn't provide significant clarity about what this function is doing different than `hubble_sat_packet_send()`. Growing function names become harder to follow when not written in spoken order. i.e. `hubble_sat_packet_send_during_pass()` reads well and is clear to me, though I know this breaks the naming convention and is probably not the preferred approach

ceolin:
Thank you @HongNguyen635 @hunterhubble @Micahscope. I have updated the API.

Micahscope:
nit: API renamed to `hubble_sat_packet_pass_send()`, PR calls out `hubble_sat_packet_send_pass()`. Not that familiar with GitHub yet, so I'm not sure if it is difficult or not to rename in the PR message

ceolin:
> nit: API renamed to `hubble_sat_packet_pass_send()`, PR calls out `hubble_sat_packet_send_pass()`. Not that familiar with GitHub yet, so I'm not sure if it is difficult or not to rename in the PR message

It does not really matter to be honest at this point. This information won't be stored in git.

HongNguyen635 (approved):

hunterhubble (approved):

PR #273, opened 2026-07-07
tests: hubble: Increase coverage of common API

ceolin:
Add new test to increase coverage of common hubble API.

### DIFF
--- a/tests/zephyr/hubble-api/src/main.c
+++ b/tests/zephyr/hubble-api/src/main.c
@@ -1,0 +1,211 @@
+/*
+ * Copyright (c) 2026 Hubble Network, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <hubble/hubble.h>
+#include <hubble/port/crypto.h>
+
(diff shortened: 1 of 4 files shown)


tests/zephyr/hubble-api/src/main.c
```diff
@@ -0,0 +1,211 @@
...
+	/* Update to a later time */
+	err = hubble_time_set(_unix_time + 60000U);
+	zassert_ok(err);
+
+	/* Time smaller than current uptime must be rejected */
+	err = hubble_time_set(1U);
+	zassert_not_ok(err);
+
+	/* Prior valid time must be preserved after the failed call */
+	zassert_not_equal(0, hubble_time_get());
```
HongNguyen635, line 135: should this check the actually time returns to see that time is preserved?

ceolin: We can't test that, because it returns uptime and we don't know how much time has elapsed. This is testing that and invalid time_set() has not corrupted the previous value.

HongNguyen635 (approved):

PR #276, opened 2026-07-08
docs: ncs: add integration guide for nRF Connect SDK

HongNguyen635:
Add integration guide for NCS, focus specifically on the Hubble dual stack and satellite board integration to the network.

See [page preview here](https://hongnguyen635.github.io/hubble-device-sdk/main/integration_guides/ncs/index.html)

### DIFF
--- a/docs/integration_guides/ncs/index.rst
+++ b/docs/integration_guides/ncs/index.rst
@@ -1,0 +1,783 @@
+.. _ncs_integration_guide:
+
+Nordic nRF Connect SDK Integration Guide
+########################################
+
+This guide walks through integrating the Hubble Network dual-stack Satellite
+and BLE application with the `nRF Connect SDK`_ (NCS).
+NCS is Nordic Semiconductor's production-ready SDK built on top of the Zephyr
+RTOS, and is the recommended platform for Nordic nRF52, nRF53, and nRF54 Series
+devices.
+
+By the end of this guide you will know how to:
+
+* Integrate the Hubble Satellite and Terrestrial (BLE) network stacks into an
+  NCS application.
+* Obtain ephemeris data and use pass prediction to schedule satellite
+  transmissions.
+* Transmit data to the Hubble satellite network from a Nordic device.
+
+Supported NCS Version and Boards
+*********************************
+
+The Hubble Network SDK currently supports **NCS v3.4.0**. If you require
+support for a different version, `contact us <mailto:support@hubble.com>`_.
+
+Nordic nRF52, nRF53, and nRF54 Series SoCs are supported. The SDK includes
+board API implementations for the following development kits out of the box:
+
+.. list-table::
+   :widths: 50 25 25
+   :header-rows: 1
+
+   * - Board
+     - Target
+     - SoC Family
+   * - nRF21540 DK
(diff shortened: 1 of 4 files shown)


docs/integration_guides/ncs/index.rst
```diff
@@ -0,0 +1,766 @@
...
+path before transmission and put it back to sleep/bypass afterwards:
+
+.. code-block:: c
+
+   int hubble_sat_board_packet_send(const struct hubble_sat_packet_frames *packet)
+   {
+       /* Enable your PA/FEM before transmitting. */
+       board_fem_enable();
+       /* Call the SoC layer to handle packet transmission */
+       hubble_sat_soc_packet_send(packet);
```
ceolin: we should return the value returned by this call

docs/integration_guides/ncs/index.rst
```diff
@@ -0,0 +1,766 @@
...
+   a firmware update is needed to refresh them.
+
+See :ref:`hubble_satellite_introduction` for details on the generated format
+and how to register the array with the SDK.
+
+For devices that need to refresh orbital parameters without a firmware update,
+the common approach is to deliver them at runtime over BLE from a companion
+application. The ``sat-dual-stack`` sample demonstrates this pattern. The
+companion app writes the parameters over a GATT characteristic and the device
+copies them into ``struct hubble_sat_orbital_params``.
```
ceolin: I think this paragraph can be removed, we have stated earlier that these parameters are static (due station keeping) so it is not clear why an application would need to refresh it.

We can revisit this topic later if things change imho.

docs/integration_guides/ncs/index.rst
```diff
@@ -0,0 +1,766 @@
...
+           return err;
+       }
+
+       /* Blocking call, retries are handled internally by the SDK */
+       err = hubble_sat_packet_send(&packet, HUBBLE_SAT_RELIABILITY_NORMAL);
+       if (err != 0) {
+           LOG_ERR("Failed to send packet (err %d)", err);
+           return err;
+       }
+   } /* end while loop, back to compute the next pass */
```
ceolin: it is worth to add to this snippet `bt_init()` and `start_adv()` as well. Or at least adding a comment where it is safe to enable bluetooth in this loop.

docs/integration_guides/ncs/index.rst
```diff
@@ -0,0 +1,766 @@
...
+
+   int hubble_sat_board_packet_send(const struct hubble_sat_packet_frames *packet)
+   {
+       /* Enable your PA/FEM before transmitting. */
+       board_fem_enable();
+       /* Call the SoC layer to handle packet transmission */
+       hubble_sat_soc_packet_send(packet);
+       /* Return PA/FEM to sleep/bypass after transmission. */
+       board_fem_sleep();
+       return 0;
```
ceolin: I think it is worth to mention that the board may (can) set the radio power here as well based on the FEM output power.

docs/integration_guides/ncs/index.rst
HongNguyen635: okay, I'll leave the TODO note there in case we want to revisit this

ceolin (approved):

PR #278, opened 2026-07-09
port: ti: Automatically setup radio

ceolin:
Add a build option that automatically sets up the radio.
If the application needs to use this FreeRTOS hook it can
disable the option but it needs to call hubble_init()
before using Bluetooth or Hubble stacks.

### DIFF
--- a/.meta/Hubble.syscfg.js
+++ b/.meta/Hubble.syscfg.js
@@ -96,6 +96,11 @@ function getOpts(mod) {
         if (hubble.dtmMode) {
             result.push(`-DCONFIG_HUBBLE_SAT_NETWORK_DTM_MODE`);
         }
+
+        if (hubble.useFreeRTOSDaemonHook) {
+            result.push(`-DconfigUSE_DAEMON_TASK_STARTUP_HOOK`);
+            result.push(`-DCONFIG_HUBBLE_FREERTOS_DAEMON_HOOK`);
+        }
     }
 
     if (hubble.useTerrestrial) {
@@ -357,6 +362,17 @@ needed to verify time-dependent behavior.
 `,
                 default: false
             },
+            {
+                name: "useFreeRTOSDaemonHook",
+                displayName: "Use FreeRTOS Daemon Hook to init radio",
+                description: `
+Use FreeRTOS daemon task startup hook to automatically configure the device
+radio. If that is not set the application must call hubble_init() before using
+Bluetooth or the Hubble APIs.
+`,
+                default: true,
+                hidden: true
+            },
         ]
     },
     templates: {
@@ -377,6 +393,7 @@ function onUseSatelliteChange(inst, ui)
 {
     ui.deviceTDR.hidden = !inst.useSatellite;
     ui.dtmMode.hidden = !inst.useSatellite;
+    ui.useFreeRTOSDaemonHook.hidden = !inst.useSatellite;
 
     // reset so a stale "true" can't leak when sat is disabled
     if (!inst.useSatellite) {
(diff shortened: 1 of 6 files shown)


.meta/Hubble.syscfg.js
```diff
@@ -357,6 +362,17 @@ needed to verify time-dependent behavior.
 `,
                 default: false
             },
+            {
+                name: "useFreeRTOSDaemonHook",
+                displayName: "Use FreeRTOS Daemon Hook to init radio",
+                description: `
+Use FreeRTOS daemon task startup hook to automatically configure the device
+radio. If that is not set the application must call hubble_init() before using
+Bluetooth or the Hubble APIs.
```
HongNguyen635, line 371: should we update this note in the [docs](https://github.com/HubbleNetwork/hubble-device-sdk/blob/c555a5482b4ef8f9e39afe58bcf679fa61ab3db2/docs/quickstart/ti/index.rst?plain=1#L155) as well?

port/freertos/boards/ti/cc23xx_cc27xx/radio.c
```diff
@@ -193,6 +178,37 @@ int hubble_sat_board_init(void)
...
 }
 
+#ifdef CONFIG_HUBBLE_FREERTOS_DAEMON_HOOK
+void vApplicationDaemonTaskStartupHook(void)
+{
+	(void)_board_init();
+}
+#endif
+
+/* TODO: make sure that the init is only called once during a lifetime of the application */
```
HongNguyen635, line 187: we can probably remove this TODO since init is only called once

port/freertos/boards/ti/cc23xx_cc27xx/radio.c
HongNguyen635: remove this todo as well since we moved the sem out of this

port/freertos/boards/ti/cc23xx_cc27xx/radio.c
```diff
@@ -193,6 +178,37 @@ int hubble_sat_board_init(void)
...
+
+		_transmit_sem = NULL;
+
+		return -EAGAIN;
+	}
+
+#ifndef CONFIG_HUBBLE_FREERTOS_DAEMON_HOOK
+	return _board_init();
+#else
+	return 0;
```
HongNguyen635: let's check for `rcl_handle`, if the `rcl_handle == NULL` this means that it was not succeeded.

port/freertos/boards/ti/cc23xx_cc27xx/radio.c
ceolin: Actually this is still valid because the radio init may not happen automatically depending on build option and the board init, semaphores creation, remains called the same way.

port/freertos/boards/ti/cc23xx_cc27xx/radio.c
ceolin: Yep, that makes sense.

.meta/Hubble.syscfg.js
ceolin: done

port/freertos/boards/ti/cc23xx_cc27xx/radio.c
```diff
@@ -193,6 +177,38 @@ int hubble_sat_board_init(void)
...
+		_transmit_sem = NULL;
+
+		return -EAGAIN;
+	}
+
+#ifndef CONFIG_HUBBLE_FREERTOS_DAEMON_HOOK
+	return _board_init();
+#else
+	/* If the radio was properly configured, we must have a valid RCL handle here. */
+	return rcl_handle != NULL: 0 ? -EIO;
```
HongNguyen635: syntax error, should be `rcl_handle != NULL ? 0 : -EIO;`

HongNguyen635 (approved):

PR #280, opened 2026-07-10
docs: improvements: Names and organization

ceolin:
- Consolidate the name of the SDK as "Hubble Device SDK"
- Put all APIs under a new section
- Rename the BLE Network to Terrestrial Network (no code changes !)
- Scattered fixes, see individual commits for more information.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,4 @@
-# Hubble Network SDK build configuration file.
+# Hubble Device SDK build configuration file.
 #
 # https://www.sphinx-doc.org/en/master/usage/configuration.html
 
@@ -8,7 +8,7 @@ from pathlib import Path
 
 # -- Project information -----------------------------------------------------
 
-project = "Hubble Network SDK"
+project = "Hubble Device SDK"
 copyright = "2025, Hubble Network, Inc"
 author = "HubbleNetwork"
 release = "0.1"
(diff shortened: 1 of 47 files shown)


docs/integration_guides/ncs/index.rst
```diff
@@ -635,7 +635,7 @@ The SDK requires a valid non-zero Unix timestamp.
 **Fix:** Ensure time is provisioned (over BLE, NTP, GPS, or RTC) before
 calling :c:func:`hubble_init`. See :ref:`ncs_sat_data_requirements`.
 
----
+----
```
HongNguyen635: when I wrote this I was kind of like how it has a line for seperating the different symptoms -> fix section, is there a way to draw a line or make each section more "separated"? Right now it looks to me like they're the same issue if skimming fast, while it should have visible break between each of the issues:

Image

docs/conf.py
```diff
@@ -42,21 +42,25 @@
...
 primary_domain = "c"
 
 # Tell sphinx what the pygments highlight language should be.
-pygments_sytle = "sphinx"
+pygments_style = "sphinx"
 highlight_language = "none"
 
+# Keep domain objects (C function/struct signatures documented by breathe) out
+# of the sidebar navigation TOC. They are still rendered in the page body.
+toc_object_entries = False
```
HongNguyen635, line 57: is this what it's supposed to look like? I'm still seeing the c function in the side bar:

Image

HongNguyen635 (dismissed):
Just some minor comments, but looks good to me overall, so no blocking merge.

docs/conf.py
ceolin: It should not, I have added it to avoid this. In my environment it is not appearing but I built using CI and I saw this happening too.

Let me look better.

docs/integration_guides/ncs/index.rst
ceolin: I can remove this one. It is fine for me.

HongNguyen635 (approved):

PR #281, opened 2026-07-11
docs: Add bare-metal quick start

ceolin:
Move the contents of https://hubble.com/docs/guides/device-integration/bare-metal to the SDK.

### DIFF
--- a/docs/quickstart/bare-metal/index.rst
+++ b/docs/quickstart/bare-metal/index.rst
@@ -1,0 +1,656 @@
+.. _bare_metal_quick_start:
+
+Bare Metal Quick Start
+######################
+
+Port the Hubble Device SDK to a bare metal environment — one without a
+traditional real-time operating system (RTOS) such as FreeRTOS or Zephyr.
+
+
+Requirements
+************
+
+- Device ``key`` (generated when you register a new device to your organization
+  through the Hubble Cloud API).
+- A **counter source** for EID rotation and key derivation — either device
+  uptime (recommended for bare metal, no clock required) or current Unix time.
+  See :ref:`bare_metal_counter_source` below.
+- Encryption library or custom implementation.
+
+
+Overview
+********
+
+Bare metal implementations run directly on the microcontroller hardware, often
+using vendor-provided Hardware Abstraction Layers (HALs) or custom timing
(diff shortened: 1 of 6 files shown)


docs/quickstart/bare-metal/index.rst
```diff
@@ -0,0 +1,656 @@
...
+   ``src/crypto/psa.c``, or provide your own implementation.
+
+
+Validation
+**********
+
+Test your port by:
+
+#. Verifying ``hubble_uptime_get()`` returns monotonically increasing values.
+#. Calling :c:func:`hubble_init` with your key and the appropriate initial value
```
HongNguyen635, line 590: since we frequently saw some cases where people miss this, just for extra redundant, we can also note that "The key must remain valid for the lifetime of SDK usage. The SDK stores the pointer directly and does not copy the key. Do not use a stack or temporary buffer."

docs/quickstart/freertos/index.rst
HongNguyen635, line 1: missing ESP-IDF [quick start](https://github.com/HubbleNetwork/hubble-device-sdk/blob/746874e67c625dde54e21e913c871eac572b6c92/docs/quickstart/esp-idf/index.rst?plain=1#L3)

HongNguyen635 (commented):
in the same commit to remove redundant text, we can probably remove the "integration guide` in this [doc](https://github.com/HubbleNetwork/hubble-device-sdk/blob/746874e67c625dde54e21e913c871eac572b6c92/docs/integration_guides/ncs/index.rst?plain=1#L3) as well?

docs/quickstart/freertos/index.rst
ceolin: All these QuickStarts are sharing the same information and are not really consistent, my plan was to copy this one as is from hubble.com/docs and than properly make these Quickstart consistent

HongNguyen635: @ceolin that makes sense, sorry I wasn't being cleared, I just mentioned that for your 2nd commit of remove redundant text, you forgot to remove the "quick start" text for ESP-IDF as well

ceolin: ohh good catch.

HongNguyen635 (approved):

PR #283, opened 2026-07-14
api: hubble_init: separate unix_time and initial_counter arguments

buckleypaul:
hubble_init() overloaded a single first argument whose meaning depended on the compile-time counter source: Unix epoch ms in Unix-time mode, or the initial EID counter in device-uptime mode. This made call sites ambiguous and, worse, made it impossible for a device-uptime deployment to also supply a Unix time, even though Unix time is independently useful there for hubble_time_get(), satellite pass prediction, and BLE advertisement timestamps.

Give unix_time and initial_counter their own dedicated arguments so each caller states its intent explicitly, and so device-uptime callers can opt into wall-clock time without changing counter-source semantics. initial_counter is a documented no-op in Unix-time mode, reserved for future use.

All in-repo callers (samples, Zephyr/unit tests, docs) are updated in this same change since this is a breaking signature change.

### DIFF
--- a/docs/integration_guides/ncs/index.rst
+++ b/docs/integration_guides/ncs/index.rst
@@ -421,7 +421,7 @@ before calling any other Hubble API:
     * At this point unix_time_ms, device_pos, and orb_params are assumed to be
     * valid. Either baked into firmware or received via BLE provisioning.
     */
-   err = hubble_init(unix_time_ms, master_key);
+   err = hubble_init(unix_time_ms, 0, master_key);
    if (err != 0) {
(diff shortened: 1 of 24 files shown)


HongNguyen635:
> This made call sites ambiguous and, worse, made it impossible for a device-uptime deployment to also supply a Unix time.

I'm thinking about this comment, if they know unix time, why would they want to use uptime counter?

I can see the benefit of this, but also see the downside is people will likely / accidentally swap these 2 args with each other. Right now they can set unix time with `hubble_time_set` though.

ceolin:
We have already discussed it and decided to not take this direction. Closing it, but feel free to re-open if anything else come up.

PR #286, opened 2026-07-23
sample: esp: sat-continuous: Added device uptime option

hunterhubble:
- Add CONFIG_HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME=y to sdkconfig.defaults to enable


ceolin (changes requested):
Please, align it with Zephyr's sample and remove the unix time path.

hunterhubble:
@ceolin The sample for Zephyr does not have the BLE time sync feature that ESP does, would you like that removed here?

HongNguyen635 (commented):
edited: you can remove the time sync file. Dual stack has it and introducing it here makes the sample a bit complicated than it should be. Also please update the `README` as well, thank you.

HongNguyen635 (approved):

ceolin:
> @ceolin The sample for Zephyr does not have the BLE time sync feature that ESP does, would you like that removed here?

Was talking about removing from this one :)

ceolin (approved):

PR #290, opened 2026-07-27
samples: sat-cont & sat-dual-stack: Add board support for nrf54l15tag and XIAO ESP32-C6

Kr0nu5:
Overlay for nrf54l15tag selects antenna 1 by default but can be switched to antenna 2 or both in the overlay, while the config sets the tx power to the max (+7dbm). The XIAO board has an smd antenna it uses by default, so a kconfig.projbuild boolean `EXTERNAL_ANTENNA` and accompanying code snippet were added to select the external antenna (u.fL) using the on-board RF switch.

### DIFF
--- a/samples/esp-idf/sat-continuous/main/main.c
+++ b/samples/esp-idf/sat-continuous/main/main.c
@@ -6,6 +6,7 @@
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
+#include "driver/gpio.h"
 
 #include "freertos/FreeRTOS.h"
 #include "freertos/task.h"
@@ -22,6 +23,38 @@
 
 static const char *APP_TAG = "sat_continuous";
 static uint8_t _hubble_key[CONFIG_HUBBLE_KEY_SIZE];
+
+// Conditional external antenna selection
+static void select_external_antenna(void)
+{
+#ifdef CONFIG_EXTERNAL_ANTENNA
+    // GPIO3: RF switch power enable (active LOW)
+    gpio_config_t io_conf3 = {
+        .pin_bit_mask = (1ULL << GPIO_NUM_3),
+        .mode = GPIO_MODE_OUTPUT,
+        .pull_up_en = GPIO_PULLUP_DISABLE,
+        .pull_down_en = GPIO_PULLDOWN_DISABLE,
+        .intr_type = GPIO_INTR_DISABLE
+    };
+    ESP_ERROR_CHECK(gpio_config(&io_conf3));
+    ESP_ERROR_CHECK(gpio_set_level(GPIO_NUM_3, 0));  // LOW = power ON
+
+    // GPIO14: Antenna select (HIGH = external)
+    gpio_config_t io_conf14 = {
+        .pin_bit_mask = (1ULL << GPIO_NUM_14),
+        .mode = GPIO_MODE_OUTPUT,
+        .pull_up_en = GPIO_PULLUP_DISABLE,
+        .pull_down_en = GPIO_PULLDOWN_DISABLE,
+        .intr_type = GPIO_INTR_DISABLE
+    };
+    ESP_ERROR_CHECK(gpio_config(&io_conf14));
+    ESP_ERROR_CHECK(gpio_set_level(GPIO_NUM_14, 1));  // HIGH = external antenna
+
+    ESP_LOGI(APP_TAG, "External antenna selected (GPIO3=LOW, GPIO14=HIGH)");
+#else
+    ESP_LOGI(APP_TAG, "Using default (onboard) antenna");
+#endif
+}
 
 void app_main(void)
 {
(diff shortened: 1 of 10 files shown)


samples/esp-idf/sat-dual-stack/main/Kconfig.projbuild
```diff
@@ -19,4 +19,10 @@ menu "Sat Dual Stack Sample Configuration"
         This will set the sat timer for 2 minutes instead
         of until a satellite pass is overhead.
 
+    config EXTERNAL_ANTENNA
```
ceolin, line 22: I don't think this is necessarily a sample an option, since it is a board property. It would be better to have it in the board Kconfig and apply it to any application using it (during the board initialization code).

Not a blocker, we can do it after ... but I think that is the right direction.

samples/esp-idf/sat-dual-stack/main/main.c
```diff
@@ -51,6 +52,38 @@ static void _sat_timer_cb(void *arg)
...
+        .intr_type = GPIO_INTR_DISABLE
+    };
+    ESP_ERROR_CHECK(gpio_config(&io_conf14));
+    ESP_ERROR_CHECK(gpio_set_level(GPIO_NUM_14, 1));  // HIGH = external antenna
+
+    ESP_LOGI(APP_TAG, "External antenna selected (GPIO3=LOW, GPIO14=HIGH)");
+#else
+    ESP_LOGI(APP_TAG, "Using default (onboard) antenna");
+#endif
+}
```
ceolin, line 85: This does not need to be duplicated in each sample if goes to the board implementation.

ceolin (changes requested):
Thanks a lot for the contribution, it is looking good. Just a few requests:
* Split the PR into two commits, one per board. It makes easier to bisect when debugging.
* Please provide a proper commit message, the body is empty, it is missing the signed-off needed by DCO. See https://github.com/HubbleNetwork/hubble-device-sdk?tab=contributing-ov-file#commit-guide
* The copyright in the file is saying Hubble, add your own.

samples/esp-idf/sat-dual-stack/main/Kconfig.projbuild
Kr0nu5: Agreed - just found the board specific dir (port/esp-idf/hubblenetwork-sdk/boards/esp32-c6), would have put the options and code there if I saw if before. I will be sure to give the board option and antenna select function better naming to make clear it is for the XIAO ESP32-C6 as well.

samples/esp-idf/sat-dual-stack/main/main.c
Kr0nu5: Same as above - found the esp32-c6 board config dir and will move the definition there.

Kr0nu5:
> Thanks a lot for the contribution, it is looking good. Just a few requests:
> 
>     * Split the PR into two commits, one per board. It makes easier to bisect when debugging.
> 
>     * Please provide a proper commit message, the body is empty, it is missing the signed-off needed by DCO. See https://github.com/HubbleNetwork/hubble-device-sdk?tab=contributing-ov-file#commit-guide
> 
>     * The copyright in the file is saying Hubble, add your own.

Thanks for the heads up - I will split the commits (with above changes) and be sure to add a proper commit message, sign-off, and copyright info.

samples/zephyr/sat-dual-stack/boards/nrf54l15tag_nrf54l15_cpuapp.overlay
```diff
@@ -0,0 +1,27 @@
...
+        line-name = "rf-switch-v1";
+    };
+
+    ant2_hog: ant2_hog {
+        gpio-hog;
+        gpios = <10 GPIO_ACTIVE_LOW>;   /* ANT2 */
+        output-low;
+        line-name = "rf-switch-v2";
+    };
+};
```
ceolin, line 27: These overlays will need to be duplicated in all builds for this target, a better approach would be define either define a snippet for this or do it programmatically in the board initialization.

samples/esp-idf/sat-continuous/main/main.c
```diff
@@ -23,6 +24,38 @@
...
+    };
+    ESP_ERROR_CHECK(gpio_config(&io_conf14));
+    ESP_ERROR_CHECK(gpio_set_level(GPIO_NUM_14, 1));  // HIGH = external antenna
+
+    ESP_LOGI(APP_TAG, "External antenna selected (GPIO3=LOW, GPIO14=HIGH)");
+#else
+    ESP_LOGI(APP_TAG, "Using default (onboard) antenna");
+#endif
+}
+
```
ceolin, line 58: I don't think this code fits in the sample, the implementation it is supporting an external antenna only in a a hardcoded gpio ...

ceolin (changes requested):
Sorry for the late reply, we have been busy with the latest release :)

Thanks for following up on it. I think the NRF54 part could be done in a snippet, but I can do it later if you prefer. The Espressif part I don't think fits in the sample.

Can you have only the nrf54 part in the commit and merge that ?

PR #293, opened 2026-07-27
hdcv: Add hubble device configuration vector

ceolin:
Add a single string (vector) representing all options set in a build. This can be used by an application to check if the fw configuration matches with the provisioned configuration.

### DIFF
--- a/include/hubble/hubble.h
+++ b/include/hubble/hubble.h
@@ -7,6 +7,7 @@
 #ifndef INCLUDE_HUBBLE_HUBBLE_H
 #define INCLUDE_HUBBLE_HUBBLE_H
 
+#include <stdbool.h>
 #include <stdint.h>
 
 #ifdef CONFIG_HUBBLE_SAT_NETWORK
@@ -126,6 +127,36 @@ int hubble_key_set(const void *key);
  * @return 0 on success, negative error code on failure (Unix time mode only, if time not set)
  */
 int hubble_counter_get(uint32_t *counter);
+
+/**
+ * @brief Get the Hubble Device Configuration Vector (HDCV) for this build.
+ *
+ * Returns a compact, fixed-format string that describes the crypto,
+ * EID-rotation and network configuration compiled into the SDK, for example
+ * `HDCV:1.0/E:256/CS:UT/RP:S86400/N:T/TV:1`. The string is derived entirely
+ * from build-time options, so it can be logged, stored, or reported to identify
+ * how a device is configured.
+ *
+ * @return Pointer to a static, NUL-terminated HDCV string. Never NULL.
+ */
+const char *hubble_config_vector_get(void);
+
+/**
+ * @brief Check whether a Hubble Device Configuration Vector (HDCV) string
+ * matches this build's configuration.
+ *
+ * Compares @p hdcv against the Hubble Device Configuration Vector for this
+ * build (the string returned by @ref hubble_config_vector_get). Use it to
+ * confirm that a device is configured as expected — for example, that the
+ * vector a provisioning tool or the Cloud recorded still matches the firmware.
+ *
+ * @param hdcv NUL-terminated HDCV string to check.
+ *
+ * @return
+ *          - true if @p hdcv is non-NULL and equals this build's HDCV string.
+ *          - false otherwise (including when @p hdcv is NULL).
+ */
+bool hubble_config_vector_check(const char *hdcv);
 
 /**
  * @}
--- a/src/utils/hdcv.h
+++ b/src/utils/hdcv.h
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2026 Hubble Network, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef SRC_UTILS_HDCV_H
+#define SRC_UTILS_HDCV_H
+
+#include "macros.h"
+
+/*
+ * Hubble Device Configuration Vector (HDCV) — a compact, fixed-format string
+ * describing this build's crypto, EID-rotation and network configuration.
+ * It is derived entirely from build options so the device can report or log
+ * its configuration.
+ *
+ *     HDCV:<version>/E:<enc>[/CS:<src>][/EC:<count>][/RP:<period>][/N:<net>][/TV:<ver>][/SV:<ver>]
+ */
+
+/* E — encryption / key size */
+#if defined(CONFIG_HUBBLE_NETWORK_KEY_256)
+#define HDCV_E "/E:256"
+#elif defined(CONFIG_HUBBLE_NETWORK_KEY_128)
+#define HDCV_E "/E:128"
+#else
+#define HDCV_E ""
+#endif
+
+/* CS — counter source  */
+#if defined(CONFIG_HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME)
+#define HDCV_CS "/CS:DU"
+#elif defined(CONFIG_HUBBLE_COUNTER_SOURCE_UNIX_TIME)
+#define HDCV_CS "/CS:UT"
+#else
+#define HDCV_CS ""
+#endif
+
+/* EC — EID count; only meaningful for device uptime, fixed at 128 today */
+#if defined(CONFIG_HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME)
+#define HDCV_EC "/EC:128"
+#else
+#define HDCV_EC ""
+#endif
+
+/* RP — rotation period in seconds */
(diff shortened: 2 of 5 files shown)


include/hubble/hubble.h
```diff
@@ -127,6 +128,36 @@ int hubble_key_set(const void *key);
...
+ * `HDCV:1.0/E:256/CS:UT/RP:S86400/N:T/TV:1`. The string is derived entirely
+ * from build-time options, so it can be logged, stored, or reported to identify
+ * how a device is configured.
+ *
+ * @return Pointer to a static, NUL-terminated HDCV string. Never NULL.
+ */
+const char *hubble_config_vector_get(void);
+
+/**
+ * @brief Check whether an Hubble Device Configuration Vector (HDCV) string
```
HongNguyen635: nit: "a Hubble Device Configuration Vector"

include/hubble/hubble.h
```diff
@@ -127,6 +128,36 @@ int hubble_key_set(const void *key);
  */
 int hubble_counter_get(uint32_t *counter);
 
+/**
+ * @brief Get the Hubble Device Configuration Vector (HDCV) for this build.
+ *
+ * Returns a compact, fixed-format string that describes the crypto,
+ * EID-rotation and network configuration compiled into the SDK, for example
+ * `HDCV:1.0/E:256/CS:UT/RP:S86400/N:T/TV:1`. The string is derived entirely
```
HongNguyen635, line 136: Should we add a docs page describing this format?

include/hubble/hubble.h
```diff
@@ -127,6 +128,36 @@ int hubble_key_set(const void *key);
...
+ * @return Pointer to a static, NUL-terminated HDCV string. Never NULL.
+ */
+const char *hubble_config_vector_get(void);
+
+/**
+ * @brief Check whether an Hubble Device Configuration Vector (HDCV) string
+ * matches this build's configuration.
+ *
+ * Compares @p hdcv against the Hubble Device Configuration Vector for this
+ * build (the string returned by hubble_config_vector_get()). Use it to confirm
```
HongNguyen635: and the `ref` directive for `@ref hubble_config_vector_get`?

src/utils/hdcv.h
```diff
@@ -0,0 +1,82 @@
...
+
+#include "macros.h"
+
+/*
+ * Hubble Device Configuration Vector (HDCV) — a compact, fixed-format string
+ * describing this build's crypto, EID-rotation and network configuration.
+ * It is derived entirely from build options so the device can report or log
+ * its configuration.
+ *
+ *     HDCV:<version>/E:<enc>[/CS:<src>][/EC:<count>][/RP:<period>][/N:<net>][/TV:<ver>][/PV:<ver>]
```
HongNguyen635: is the last one meant to be `SV` instead of `PV`?

src/utils/hdcv.h
```diff
@@ -0,0 +1,82 @@
...
+#endif
+
+/* EC — EID count; only meaningful for device uptime, fixed at 128 today */
+#if defined(CONFIG_HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME)
+#define HDCV_EC "/EC:128"
+#else
+#define HDCV_EC ""
+#endif
+
+/* RP — rotation period in seconds (omitted when at the 86400 default) */
```
HongNguyen635: kind of contradict with the RFC:
> All other fields — E, CS, RP, N — are always present.

So either make a change to the RFC or make RP always present.

src/utils/hdcv.h
```diff
@@ -0,0 +1,82 @@
...
+/* E — encryption / key size */
+#if defined(CONFIG_HUBBLE_NETWORK_KEY_256)
+#define HDCV_E "/E:256"
+#elif defined(CONFIG_HUBBLE_NETWORK_KEY_128)
+#define HDCV_E "/E:128"
+#else
+#define HDCV_E ""
+#endif
+
+/* CS — counter source (UNIX_TIME is the default and is omitted) */
```
HongNguyen635: same thing here, the RFC states that `CS` is always present. I think we shouldn't omitted it though, could make it ambiguous later down the road. Besides, we now need to check what this empty string correspond to which default, and the default could change in future version.
We can remove the `(UNIX_TIME is the default and is omitted)` comment, what do you think?

src/utils/hdcv.h
```diff
@@ -0,0 +1,82 @@
...
+ * is the only version in this port) */
+#if defined(CONFIG_HUBBLE_BLE_NETWORK)
+#define HDCV_TV "/TV:0"
+#else
+#define HDCV_TV ""
+#endif
+
+/* PV — protocol version (satellite V1 only; terrestrial v0 is implicit) */
+#if defined(CONFIG_HUBBLE_SAT_NETWORK_PROTOCOL_V1)
+#define HDCV_SV "/SV:0"
```
HongNguyen635, line 73: This is fine for now, but we could have PHY v1 but with different payload version. Is there a better way do get the version instead of hardcode like this?

src/utils/hdcv.h
```diff
@@ -0,0 +1,82 @@
...
+#else
+#define HDCV_SV ""
+#endif
+
+#define HUBBLE_DEVICE_CONFIG_VECTOR                                            \
+	"HDCV:1.0" HDCV_E HDCV_CS HDCV_EC HDCV_RP HDCV_N HDCV_TV HDCV_SV
```
HongNguyen635, line 79: I'd prefer not to hardcode the HDCV version here, but it could be fine for this 1st version. Maybe make it a macro for now?

include/hubble/hubble.h
ceolin: Possibly but not mandatory I think, this is a contract between our infra and endpoint.

src/utils/hdcv.h
ceolin: yep :)

src/utils/hdcv.h
ceolin: Yep, I forgot update the doc. RP is always defined, unless CONFIG_HUBBLE_EID_ROTATION_PERIOD_SEC is not defined and in this case it means we have a problem and makes sense not having it.

src/utils/hdcv.h
ceolin: It is always defined, the documentation is stale because I did it before updating the RFC.

src/utils/hdcv.h
ceolin: The physical is not part of the provisioning, if it happens to need it we bump the HDCV version.

src/utils/hdcv.h
ceolin: It is hardcoded, either in a separated macro or here :) I can make it a separated macro though.

include/hubble/hubble.h
HongNguyen635: @ceolin but it'd be nice to have maybe a table document this so people can reference it across different versions. For example, they can see the string and understand what it is, and they don't have to reach out for support to understand it.

src/utils/hdcv.h
HongNguyen635: Would we ever have this scenario:

HDCV v1

with 2 possible protocol version:

v0 - current protocol
v1 - some placeholder protocol

Current we hardcoded `SV:0`, so what if we want to support `SV:1`? I assume by that time you'll pull in the kconfig symbol and ifdef like other fields?

src/utils/hdcv.h
HongNguyen635: this is fine, just want to point out.

src/utils/hdcv.h
ceolin: Yes, if we have more payload versions for satellite than we conditionally set it as it is done for other symbols, for now does not make sense since there is only one.

Regarding versioning the satellite physical version, we need to see how it will be, and how it will work for customers (whether or not they will choose which one ... ). Then if we need we bump the HDCV version and add a new field for that.

include/hubble/hubble.h
ceolin: I have a tool and a documentation that I didn't push because we need to have some sort of organized troubleshooting section where it will make sense imho

ceolin: It can come after this

HongNguyen635: Ok sounds good

src/utils/hdcv.h
HongNguyen635: that makes sense. Originally, I was just thinking about multiple payload protocol versions. I don't think we should worry about PHY version because iirc sat can only do 1 protocol at a time. But yeah we'll need to bump the HDCV if that happens.

HongNguyen635 (approved):

PR #295, opened 2026-07-29
docs: add a version selector dropdown

HongNguyen635:
Add version selector drowdown + add pycache to `.gitignore`.

The dropdown looks like this (bottom of nav bar):

Screenshot 2026-07-29 at 12 46 05 PM

I'll manually generate the docs for v1.0 and v2.0 and push to gh-pages later so those pages can have the version dropdown as well.

### DIFF
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -139,3 +139,30 @@ jobs:
           destination_dir: ${{ needs.build.outputs.slug }}
           keep_files: true
           commit_message: "Deploy docs for ${{ needs.build.outputs.slug }}"
+
+      # Check out gh-pages so we can update the versions file
+      - name: Check out gh-pages
+        uses: actions/checkout@v4
+        with:
+          ref: gh-pages
+          path: gh-pages
+
+      - name: Update Versions File
+        working-directory: gh-pages
+        env:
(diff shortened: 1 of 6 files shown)


.github/workflows/build-docs.yaml
```diff
@@ -139,3 +139,30 @@ jobs:
...
+          # and write to versions.json
+          gh api "repos/$GITHUB_REPOSITORY/branches" --paginate --jq '.[].name' \
+            | grep -E '^main$|^release/' \
+            | jq -R . | jq -s . > versions.json
+          git add versions.json
+          if git diff --cached --quiet; then
+            echo "versions.json unchanged - nothing to commit"
+          else
+            git config user.name  "github-actions[bot]"
+            git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
ceolin, line 165: are these user credentials recommended / defined anywhere ? if not can't we make them secrets / repo configuration ?

HongNguyen635: There isn't an official guide, but it's like the default whenever someone wants to commit in the action. I can use an action from a market place, it'll use the same bot, just has extra setup steps. Here I just do a simple commit, so I thought I can specify it here.
This is where I reference this info: https://github.com/orgs/community/discussions/119597

ceolin (approved):

PR #296, opened 2026-07-30
github: actions: Add codeql action

ceolin:
Add an action to run static analyais.

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -1,0 +1,159 @@
+# Copyright (c) 2026 Hubble Network, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+name: CodeQL
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+concurrency:
+  group: codeql-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'push' }}
+
+jobs:
+  cpp-zephyr:
+    name: Analyze C/C++ (Zephyr samples + tests)
+    runs-on: ubuntu-24.04
+    timeout-minutes: 360
+    permissions:
+      contents: read
+      actions: read
+      security-events: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+          cache: pip
+
+      - name: Setup Zephyr project (SDK + modules)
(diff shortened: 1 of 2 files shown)


ceolin:
example of the analysis https://github.com/ceolin/hubble-device-sdk/security/code-scanning

.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,173 @@
...
+    branches:
+      - main
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+concurrency:
+  group: codeql-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
HongNguyen635: we restrict this to push on `main`, so the event pull request would never be true. Is the intention of this block is when there is a new push to main, we cancel in progress run and run the new one? If so, I think we can just make it like this:

`cancel-in-progress: true`

.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,173 @@
...
+        with:
+          category: "/language:c-cpp/build:zephyr"
+          output: sarif-results
+          upload: failure-only
+
+      - name: Upload results
+        uses: github/codeql-action/upload-sarif@v4
+        with:
+          sarif_file: sarif-results
+          category: "/language:c-cpp/build:zephyr"
```
HongNguyen635: why is this split into 2 steps? Can we upload it on the analysis step as well or is it because of the memory issue?

.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,173 @@
...
+            --recursive \
+            --target esp32c6 \
+            --manifest-file tools/ci/esp-idf-build-manifest.yaml
+
+      - name: Perform CodeQL analysis
+        uses: github/codeql-action/analyze@v4
+        with:
+          category: "/language:c-cpp/build:esp-idf"
+          output: sarif-results
+          upload: failure-only
```
HongNguyen635: same here

.github/workflows/codeql.yaml
ceolin: I had initially add to pr and later limited it. Will remove it.

.github/workflows/codeql.yaml
ceolin: different actions

HongNguyen635: because I read [here](https://github.com/github/codeql-action/blob/5d3eb98e4a780bfe1a53f57fbdc278ea5da1274b/analyze/action.yml#L15) and it said the default is always upload unless you need to process things inbetween, that is when we need to split.

ceolin: Ohh I missed it. It makes sense.

HongNguyen635 (approved):

PR #298, opened 2026-07-30
port: zephyr: nrf52: manage HFCLK via the onoff manager

ceolin:
The satellite SoC enable path turned the high-frequency clock on with clock_control_on() but never turned it off — hubble_sat_soc_disable() returned 0 without releasing it — so the HFCLK stayed running after a satellite transmission and the request never participated in Zephyr's clock reference counting.

Different components on Zephyr that use this clock do it through onoff manager and if we try to use it differently we get error due different hw ownership. This fix the problem.


HongNguyen635 (commented):
Should we edit nrf53 / 54 as well? Since they also call `clock_control_on`?

ceolin:
@HongNguyen635 I thought about it, but I am avoiding changing critical things now unless strictly necessary. Since, this code path is not used with SoftDevice I didn't change it now.

HongNguyen635 (approved):

hunterhubble (approved):
Validated on a Seeed XIAO nrf52840. Thank you!

PR #306, opened 2026-07-31
docs: release: Add WIP v3.0.0 release notes

ceolin:
Add working in progress release notes for v3.0.0.

### DIFF
--- a/docs/releases/3.0.rst
+++ b/docs/releases/3.0.rst
@@ -1,0 +1,175 @@
+.. _release_3_0_0:
+
+Release 3.0.0
+#############
+
+**Status**: Release Candidate (``v3.0.0-rc1``)
+
+**Release Date**: TBD
+
+Release 3.0.0 is a major release focused on the **Satellite Network**. It
+matures the satellite pass-prediction pipeline, introduces a Direct Test Mode
+(DTM) for RF bring-up, and adds broad new platform and board support across
+Zephyr, FreeRTOS/TI, and ESP-IDF. With this release, the Satellite Network
+graduates to a production-ready, stable API, which drives the MAJOR version
+increment.
+
+Overview
+********
+
+The headline of 3.0.0 is a substantially more capable and correct Satellite
+Network. Pass prediction now reports the maximum elevation angle, pass start
+time, and pass duration, and lets applications filter passes by a configurable
+minimum elevation angle. A new Direct Test Mode exposes low-level transmit and
+continuous-wave controls for RF validation and certification. Satellite radio
+support has been added for TI CC27xx/CC23xx, Nordic nRF52/nRF53/nRF54 (including a
+SoftDevice-based path), Silicon Labs EFR32xG24, and Espressif ESP32-C6, and a
+full set of dual-stack, continuous, and DTM samples ships for each platform.
+
+As part of stabilizing the API, the pass-prediction functions were moved into
+the ``hubble_sat_`` namespace and now use milliseconds for all time values,
+aligning them with the rest of the SDK.
+
+Key Features
+************
+
+Satellite Pass Prediction
+=========================
+
+- Pass results now report richer information via ``struct hubble_sat_pass_info``:
+  the maximum elevation angle (``max_elevation_angle``), the pass start time
+  (``start``), and the pass duration (``duration``).
+- :c:func:`hubble_sat_min_elevation_angle_set` filters out passes whose peak
+  elevation is below a caller-defined angle, so applications only transmit when a
+  satellite is high enough in the sky.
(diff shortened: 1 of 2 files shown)


docs/releases/3.0.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Sample Applications
+*******************
+
+New samples shipping with 3.0.0:
+
+- **ESP-IDF**: ``sat-continuous``, ``sat-dtm``, and ``sat-dual-stack``.
+- **Zephyr**: ``sat-dual-stack``, ``sat-dtm``, ``sat-continuous`` (with board
+  overlays, including Silicon Labs xG24 dev boards), and ``ble-beacon-mfg-data``.
+- **FreeRTOS / TI**: ``ble-beacon``, ``sat-continuous``, ``sat-dtm``, and
```
HongNguyen635: `ble-beacon` is not a new sample right?

docs/releases/3.0.rst
```diff
@@ -0,0 +1,159 @@
...
+  ``sat-dual-stack``.
+
+Documentation
+=============
```
HongNguyen635: Is Documentation supposed to be nested under `Sample Applications`? If not we can make it `***...` instead of `===...`

docs/releases/3.0.rst
```diff
@@ -0,0 +1,159 @@
...
+- Terrestrial Network overview and clarified BLE/terrestrial naming.
+
+API Highlights
+**************
+
+Satellite Network APIs:
+
+- :c:func:`hubble_sat_packet_get` — build a packet from payload bytes
+- :c:func:`hubble_sat_packet_frames_get` — split a packet into radio frames
+- :c:func:`hubble_sat_packet_send` — transmit a packet via satellite
```
HongNguyen635: should this be "to satellite" instead of "via satellite"?

docs/releases/3.0.rst
```diff
@@ -0,0 +1,159 @@
...
+Testing
+*******
+
+This release has been validated through:
+
+- Unit tests for BLE advertisement generation and encryption
+- Unit tests for EID generation in both counter source modes
+- Satellite packet encoding, frame-splitting, and payload-length validation tests
+- Pass-prediction (ephemeris) unit tests
+- Platform-specific CI builds: Zephyr (QEMU, native_sim), FreeRTOS (TI), ESP-IDF
```
HongNguyen635: Add NCS as well (although we can kind of count it as Zephyr)?

docs/releases/3.0.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Support
+*******
+
+- **Documentation**: https://hubble.com/docs/device-sdk/intro
```
HongNguyen635: I think this page has been removed. Do you want to point it to our SDK GH page instead?

https://hubblenetwork.github.io/hubble-device-sdk/release-3.0-branch/

^ This doesn't exist yet but will redirect to the right link once we create the branch.

docs/releases/3.0.rst
HongNguyen635, line 1: btw should we add a migration guide? Since it's a major release.

ceolin: I don't think we have any API change affecting Terrestrial network and the Satellite network was pre-production / not supported.

Do you have anything particular in mind ?

HongNguyen635: oh you're right, I forgot about the pre-production API not supported. I don't see the HDCV in the release note, should we mention it as well?

docs/releases/3.0.rst
ceolin: depends whether or not you consider satellites proxy :)

docs/releases/3.0.rst
ceolin: Actually I will drop this section, we are already in the documentation it does not need to add this recursion :)

docs/releases/3.0.rst
```diff
@@ -0,0 +1,151 @@
...
+
+New samples shipping with 3.0.0:
+
+- **ESP-IDF**: ``sat-continuous``, ``sat-dtm``, and ``sat-dual-stack``.
+- **Zephyr**: ``sat-dual-stack``, ``sat-dtm``, ``sat-continuous`` (with board
+  overlays, including Silicon Labs xG24 dev boards), and ``ble-beacon-mfg-data``.
+- **FreeRTOS / TI**: ``sat-continuous``, ``sat-dtm``, and ``sat-dual-stack``.
+
+Documentation
+*****************
```
HongNguyen635: extra `*`

docs/releases/3.0.rst
ceolin: added :)

HongNguyen635 (approved):

PR #308, opened 2026-08-03
FIX: Fix regression in satellite packet fix

ceolin:
- Fix regression introduced in f329e45
- Add an unit test to avoid it from happening in future


tests/zephyr/unit/sat-packet/src/test_vectors.c
```diff
@@ -0,0 +1,89 @@
...
+ * period, which is what tests/zephyr/unit/sat-packet/prj.conf configures.
+ */
+
+#include "test_vectors.h"
+
+const uint8_t test_payload[13] = {0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06,
+				  0x07, 0x08, 0x09, 0x0a, 0x0b, 0x0c};
+
+/* Same key as the BLE advertise unit test, for consistency. */
+const uint8_t test_key[32] = {0xcd, 0x15, 0xa5, 0xab, 0xc0, 0x60, 0xb6, 0x72,
```
HongNguyen635, line 19: Do we want to use the `HUBBLE_SAT_PAYLOAD_MAX` and `CONFIG_HUBBLE_KEY_SIZE` here instead?

tests/zephyr/unit/sat-packet/src/test_vectors.h
```diff
@@ -0,0 +1,36 @@
...
+	size_t payload_len;      /* Payload length (0, 4, 9 or 13) */
+	size_t expected_len;     /* Expected packet->length, in symbols */
+	uint8_t expected[HUBBLE_PACKET_MAX_SIZE]; /* Expected packet->data */
+};
+
+/* Cleartext payload; the shorter cases use a prefix of this buffer. */
+extern const uint8_t test_payload[13];
+
+/* Master key the vectors were generated with. */
+extern const uint8_t test_key[32];
```
HongNguyen635, line 31: same here

HongNguyen635 (commented):
Minor comments, but it's good that you found out about this before the release.

tests/zephyr/unit/sat-packet/src/test_vectors.c
ceolin: Not really, because this is independent of the configuration. They key is hardcoded to this size.

tests/zephyr/unit/sat-packet/src/test_vectors.h
ceolin: ditto

HongNguyen635 (approved):

hunterhubble (approved):
Tested with 13b payloads

PR #310, opened 2026-08-06
Release 3.0

ceolin:
- Small tweaks in documentation
- Bump version to 3.0.0

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -13,6 +13,22 @@ Network. Designed for developers who want a straightforward and reliable way to
 harness Hubble’s powerful capabilities, this SDK provides a suite of tools and
 libraries that make it easy to interact with all of Hubble’s services.
 
+Create your Hubble Account
+**************************
+
+To register a device on the network, you'll need a Hubble account — sign up for
+free at `dash.hubble.com <https://dash.hubble.com/create-account>`_ if you don't
+already have one.
+
+From there you can `register a device
+<https://hubble.com/docs/guides/cloud-integration/register-devices>`_ and start
+receiving its data over the cloud.
+
+* In the Dashboard: Use "Add a Device" to generate a device key.
+* Via API: Use the `Cloud API
+  <https://hubble.com/docs/api-specification/hubble-platform-api>`_, with an API
+  token and Org ID from your Dashboard.
+
 Using Hubble Device SDK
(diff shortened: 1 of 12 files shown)


docs/satellite/index.rst
```diff
@@ -166,8 +166,9 @@ a C source file with hard-coded satellite parameters and build it into the
 application. The ``tools/orbital_params_fetch.py`` helper fetches current
 orbital parameters from the Hubble API and writes ``sat_params.c``.
 
-Set ``HUBBLE_API_TOKEN`` to a Hubble API bearer token, then run the tool from the
-SDK repository:
+Set ``HUBBLE_API_TOKEN`` to a Hubble API bearer token (create one in the
+`Hubble Dashboard <https://dash.hubble.com>`_), then run the tool from the SDK
```
HongNguyen635: maybe a direct link to where to create token would help?
https://dash.hubble.com/developer/api-tokens

docs/releases/3.0.rst
```diff
@@ -3,10 +3,6 @@
 Release 3.0.0
 #############
 
-**Status**: Release Candidate (``v3.0.0-rc1``)
-
-**Release Date**: TBD
```
HongNguyen635, line 8: why don't we include this info like the other release?

docs/index.rst
HongNguyen635, line 56: Do you want to add a 6th card link to Hubble Dashboard like @danielr-hubblenetwork suggested?

ceolin: I don't have a proper icon / visual for that and we have added a section before this.

docs/releases/3.0.rst
ceolin: I can add it back. The reason was just because this information will be associated directly with the release tag.

docs/index.rst
danielr-hubblenetwork: If you could find an icon, I think it's worth adding. But I don't die for this.

HongNguyen635: Maybe this could work? https://fontawesome.com/v4/icon/th-large
Not a big deal if not added since we have a section before this.

ceolin: image

HongNguyen635 (approved):
LGTM

PR #311, opened 2026-08-07
esp-idf: add SDK to ESP Component Registry

HongNguyen635:
Add necessary files to register the Hubble Network SDK as an ESP Component for the Component Registry. This includes a CMakeLists.txt, Kconfig, and idf_component.yml file.

I uploaded a test version on the staging server here https://components-staging.espressif.com/components/hongnguyen635/hubble-device-sdk/versions/3.0.0

I also downloaded the files and go through the flow to make sure that it works. Will add instruction on how to use and CI to auto upload on release later.

### DIFF
--- a/idf_component.yml
+++ b/idf_component.yml
@@ -1,0 +1,21 @@
+description: "On-device SDK for connecting to Hubble's Terrestrial (Bluetooth Low Energy) and Satellite Networks, built on a modular architecture that ports across RTOSes and hardware."
+tags:
+  - "Bluetooth"
+  - "BLE"
+  - "Satellite"
+files:
+  use_gitignore: true
+  exclude:
+    - ".git/**/*"
+    - ".github/**/*"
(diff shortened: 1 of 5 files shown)


idf_component.yml
```diff
@@ -0,0 +1,21 @@
...
+tags:
+  - "Bluetooth"
+  - "BLE"
+  - "Satellite"
+files:
+  use_gitignore: true
+  exclude:
+    - ".git/**/*"
+    - ".github/**/*"
+    - ".venv/**/*"
```
ceolin, line 11: Wasn't this already added to .gitignore ?

ceolin (approved):

idf_component.yml
HongNguyen635: Yes but for some reason, when I run ESP tool to pack and upload, it still pulled in a couple of stuff in .venv, so I thought just put it here to be safe.

PR #314, opened 2026-08-11
zephyr: Kconfig: Declare SILABS_SISDK_RAIL_MULTIPROTOCOL

buckleypaul:
Kconfig.silabs sets a configdefault on SILABS_SISDK_RAIL_MULTIPROTOCOL, which Zephyr declares in modules/hal_silabs/simplicity_sdk/Kconfig. That glue only exists in trees carrying the Simplicity SDK HAL, so on older trees the configdefault is the symbol's only definition. Kconfig then reports it as typeless and Zephyr turns that warning into an error, failing every build at CMake configure:

  warning: SILABS_SISDK_RAIL_MULTIPROTOCOL (defined at
    port/zephyr/Kconfig.silabs:11) defined without a type
  error: Aborting due to Kconfig warnings

This broke all TI board builds on the simplelink-zephyr v3.7.0 fork. SOC_GECKO_CUSTOM_RADIO_PHY is unaffected: soc/silabs/Kconfig declares it in both old and new trees.

Declare the symbol here so it always has a type. Where the real definition exists, Kconfig merges the two and its prompt and dependencies still apply; where it does not, the symbol is inert because nothing references it.

Verified with both the v3.7.0-ti and v4.4.0 kconfiglib against trees with and without the upstream declaration: the warning is gone in all four combinations and the resulting value is unchanged.

### DIFF
--- a/port/zephyr/Kconfig.silabs
+++ b/port/zephyr/Kconfig.silabs
@@ -1,6 +1,17 @@
 # Copyright (c) 2026 Hubble Network, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
+# Zephyr declares SILABS_SISDK_RAIL_MULTIPROTOCOL in
+# modules/hal_silabs/simplicity_sdk/Kconfig, which only exists in trees that
+# carry the Simplicity SDK HAL glue. On older trees (for example the TI
+# simplelink-zephyr v3.7.0 fork) the configdefault below is the symbol's only
+# definition, so Kconfig reports it as typeless and the build aborts on the
+# warning. Declaring it here gives it a type everywhere; where the real
+# definition exists, Kconfig merges the two and its prompt and dependencies
+# still apply.
+config SILABS_SISDK_RAIL_MULTIPROTOCOL
+	bool
+
 if HUBBLE_SAT_NETWORK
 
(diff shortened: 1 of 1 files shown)


ceolin:
I tried to do something else to avoid re-define the symbol because it weakens possible symbol dependencies but there is no proper way.

The problem is not only for TI, it also affects Zephyr's 3.7.

ceolin (approved):

HongNguyen635:
@ceolin is this just an issue with zephyr 3.7? Because we can argue the same case for Nordic right? If someone using SiLabs zephyr fork for example, they don't pull in hal_nordic, would it cause the same error then?

ceolin:
> @ceolin is this just an issue with zephyr 3.7? Because we can argue the same case for Nordic right? If someone using SiLabs zephyr fork for example, they don't pull in hal_nordic, would it cause the same error then?

Yes, that is a problem with Zephyr's version. And yes, you are right, but the problem is that we can't conditionally include these files and the way Kconfig works is that you have to have the symbols defined. The problem with default implementation is that we may weakening the symbol dependencies, so we should avoid as much ash possible.

I think how it is now is a reasonable tradeoff to be honest. It is hard to promise compatibility with any downstream trees

PR #316, opened 2026-08-12
samples: zephyr/ble-beacon: Fix build for older Zephyr

ceolin:
Fix the build for older Zephyr versions.


samples/zephyr/ble-beacon/src/store.c
HongNguyen635: do we also do the same check to use `FIXED_PARTITION_DEVICE` as well?

HongNguyen635 (approved):

PR #320, opened 2026-08-18
zephyr: nrf: Use mpsl fem with softdevice

ceolin:
Use mpsl fem when building using nrf-sdk + softdevice controller.


hunterhubble (approved):
Tested with the sat-dual-stack sample on the debug mode on the nrf21540dk, received BLE packets and satellite packets.

HongNguyen635 (approved):
+1 to confirm this is working.

PR #322, opened 2026-08-19
ci: add Zephyr LTS test workflow

ceolin:
Add west-zephyr-lts.yml pinning Zephyr to v3.7-branch and a new zephyr-lts-tests.yaml workflow to build and run tests/samples against it, using a separate quarantine list that excludes PSA crypto scenarios not supported on that release.

### DIFF
--- a/.github/workflows/zephyr-lts-tests.yaml
+++ b/.github/workflows/zephyr-lts-tests.yaml
@@ -1,0 +1,95 @@
+name: Build and Run Zephyr LTS (v3.7) tests
+
+on:
+  pull_request:
+    branches: ["**"]
+    paths:
+      - "include/**"
+      - "port/zephyr/**"
+      - "src/**"
+      - "samples/zephyr/**"
+      - "tests/zephyr/**"
+      - "west-zephyr-lts.yml"
+      - "tools/ci/twister-quarantine-lts.yaml"
+      - ".github/workflows/zephyr-lts-tests.yaml"
+
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: "Optional PR number to run against (use this to target a PR head)"
+        required: false
+      commit_sha:
+        description: "Optional commit SHA to checkout (overrides pr_number if provided)"
+        required: false
+      ref:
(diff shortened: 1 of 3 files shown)


.github/workflows/zephyr-lts-tests.yaml
```diff
@@ -0,0 +1,94 @@
+name: Build and Run Zephyr LTS (v3.7) tests
+
+on:
+  pull_request:
+    branches: ["**"]
+    paths:
```
Micahscope, line 6: Does port/zephyr/** need to be added here?

ceolin: yes sir :)

.github/workflows/zephyr-lts-tests.yaml
```diff
@@ -0,0 +1,95 @@
...
+          # instead of as siblings of it.
+          app-path: app
+          manifest-file-name: west-zephyr-lts.yml
+          sdk-version: 0.17.0
+
+      - name: Run tests
+        working-directory: app
+        run: |
+          set -euxo pipefail
+          west twister -T tests/zephyr --quarantine-list tools/ci/twister-quarantine-lts.yaml
```
Micahscope, line 95: Should this be running ` west twister -T samples/zephyr <ARGS>` as well, to ensure sample apps don't get broken for Zephyr LTS?

HongNguyen635: +1 to if we should run sample as well. Although I think some of the samples will break.

HongNguyen635 (commented):
Is this LTS just aiming for the terrestrial network or sat as well? Aiming to support backward compatibility with sat on 3.7 LTS is going to be complicated as there was some API changes between RAIL iirc.

.github/workflows/zephyr-lts-tests.yaml
ceolin: So, that is the thing, this is mostly to check anything broke in LTS which should be covered by tests, but to be fair build a couple of additional apps won't hurt, most of the cost / time goes into setup the environment.

HongNguyen635 (approved):

PR #323, opened 2026-08-21
tests: ci: zephyr: Increase test coverage

ceolin:
Increase test coverage building dual stack and dtm tests for supported hw.


samples/zephyr/sat-dual-stack/sample.yaml
```diff
@@ -6,39 +6,43 @@ sample:
...
       - CONFIG_HUBBLE_NETWORK_CRYPTO_MBEDTLS=y
       - CONFIG_HUBBLE_NETWORK_KEY_128=y
-      - CONFIG_SAMPLE_PROVIDE_SAT_BOARD_SUPPORT=y
   app.sat.dual_stack.ncs:
     extra_configs:
       - CONFIG_SAMPLE_PROVIDE_SAT_BOARD_SUPPORT=n
     modules:
       - nrf
+    tags:
+      - ncs
```
HongNguyen635, line 45: is this redundant? Since I already filtered it with `modules: - nrf`

port/zephyr/boards/nordic/sat_dtm.c
```diff
@@ -23,14 +23,17 @@
 
 #ifndef RADIO_TXPOWER_TXPOWER_Pos3dBm
 #define RADIO_TXPOWER_TXPOWER_Pos3dBm (0x03UL)
+#define NRF_RADIO_TXPOWER_POS3DBM     RADIO_TXPOWER_TXPOWER_Pos3dBm
```
HongNguyen635, line 26: that's a bit odd because it should have been defined based on the condition [here](https://github.com/zephyrproject-rtos/hal_nordic/blob/4387c79cebd31927fb1ea7d64bee11728ae8041f/nrfx/hal/nrf_radio.h#L704). Is it because of the link / build order?

samples/zephyr/sat-dual-stack/sample.yaml
ceolin: Nope, that is used in CI by NCS build to exclude the other tests now.

So `modules -nrf` ensure that Zephyr's ci don't build it and this tag ensures that NCS's ci don't build the others (they don't use the snippet)

port/zephyr/boards/nordic/sat_dtm.c
ceolin: inclusion path precedence :) That header is included before we have defined `RADIO_TXPOWER_TXPOWER_Pos3dBm`

samples/zephyr/sat-dual-stack/sample.yaml
HongNguyen635: ah okay that makes sense now

HongNguyen635 (approved):

PR #324, opened 2026-08-25
ci: fix footprint summary for sysbuild

HongNguyen635:
For target that enables sysbuild, twister will error out because there is no `footprint` at the top-level twister output dir. For the fix, we need to recursively search, generate, and summarize footprint for each domain manually.


ceolin (approved):
It seems ok

PR #326, opened 2026-08-26
ci: update CI trigger to include path that affects each CI tests

HongNguyen635:
Update the CI build for each port to include paths to source code that would benefit from a CI rebuild to check for issues.

### DIFF
--- a/.github/workflows/build-samples-esp-idf.yaml
+++ b/.github/workflows/build-samples-esp-idf.yaml
@@ -5,8 +5,11 @@ on:
     branches: ["**"]
     paths:
       - "include/**"
+      - "port/esp-idf/**"
+      - "port/freertos/**"
       - "src/**"
       - "samples/esp-idf/**"
+      - "tools/ci/esp-idf-build-manifest.yaml"
       - ".github/workflows/build-samples-esp-idf.yaml"
 
   workflow_dispatch: {}
--- a/.github/workflows/ncs-tests.yaml
+++ b/.github/workflows/ncs-tests.yaml
@@ -5,9 +5,13 @@ on:
     branches: ["**"]
     paths:
       - "include/**"
+      - "port/zephyr/**"
       - "src/**"
       - "samples/zephyr/**"
       - "tests/zephyr/**"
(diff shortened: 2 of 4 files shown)


.github/workflows/build-samples-esp-idf.yaml
```diff
@@ -5,8 +5,11 @@ on:
     branches: ["**"]
     paths:
       - "include/**"
+      - "port/esp-idf/**"
+      - "port/freertos/**"
```
Micahscope, line 9: Is gating on freertos port changes intentional?

.github/workflows/ncs-tests.yaml
```diff
@@ -5,9 +5,11 @@ on:
     branches: ["**"]
     paths:
       - "include/**"
+      - "port/zephyr/**"
       - "src/**"
       - "samples/zephyr/**"
       - "tests/zephyr/**"
+      - "west-ncs.yml"
```
Micahscope, line 13: Should we also gate on changes to zephyr/module.yml (in case blobs change)?

I do see `run: west blobs fetch hubblenetwork-sdk` done in the build, so it isn't strictly necessary, but it seems good to be explicit

Since zephyr-tests.yaml adds tools/ci/twister-quarantine.yaml, I would here too, for continuity. Like the above note, I think it isn't strictly necessary. i.e. a new build is quarantined because it broke due to a change in src/, the change in src/ would gate the build, and the build skips anything in the updated quarantine

.github/workflows/build-samples-esp-idf.yaml
HongNguyen635: yeah ESP-IDF is Espressif-flavored of FreeRTOS

.github/workflows/ncs-tests.yaml
HongNguyen635: Good point I'll include the module.yml. For the quarantine, that was my initial thought as well, although I can add it for consistency with vanilla zephyr tests.

ceolin (approved):

PR #328, opened 2026-08-28
sat: esp-idf: remove rftest dependency

HongNguyen635:
Removing `rftest` dependency per Espressif recommendation.
For Hubble sat transmission to work, `esp_phy_enable()` must be called, which can be achieved by calling rftest or bt init. However, having rftest init cal at the same time as bt can cause issue for other SoCs.

### DIFF
--- a/samples/esp-idf/sat-continuous/main/main.c
+++ b/samples/esp-idf/sat-continuous/main/main.c
@@ -12,6 +12,8 @@
 
 #include "esp_log.h"
 #include "esp_system.h"
+#include "nvs_flash.h"
+#include "nimble/nimble_port.h"
 
 #include "mbedtls/base64.h"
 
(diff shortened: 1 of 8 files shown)


ceolin (dismissed):

samples/esp-idf/sat-continuous/main/main.c
```diff
@@ -46,6 +57,13 @@ void app_main(void)
 					 ESP_LOG_DEBUG);
 	}
 
+	/* Init Bluetooth (NimBLE) */
```
Micahscope: Just wondering - was there a method to the order of initialization here?

I'm not too concerned about it, just asking. I would imagine NVS must be initialized before NimBLE based on my understanding of this change.

In a prior role, on complex devices, I have seen large `pwrp` / `init` / `pwrdn` sequencing become unwieldy, and when not well documented, can lead to crashes when things are accidentally placed out of order (when adding new features / reordering for power optimization, for examples of when order is changed).

I think when we have ordering dependencies in these sequences, it is good practice to document w/ comments that X must happen before Y.

As a nit, keeping init order as close as possible between `sat-continuous` and `sat-dtm` is probably preferred

HongNguyen635: That's right, nvs is init before BT, as outlined from [esp-idf tutorial / example](https://github.com/espressif/esp-idf/blob/fff9895c82d744c7237be8847347bdd1b07c6643/examples/bluetooth/nimble/bleprph/tutorial/bleprph_walkthrough.md?plain=1#L89). I also put the order here in the [integration guide](https://hubblenetwork.github.io/hubble-device-sdk/main/integration_guides/esp-idf/index.html#initializing-the-hubble-device-sdk) (below the warning box).

For sat-cont and sat-dtm order, what is the difference you see in the init order? Sat-cont needs the key but sat-dtm doesn't need it.

Micahscope: Sorry, to clarify, nothing seems "in the wrong order", I would just expect nimble init right after NVS init, with device key decrypted after nimble init. Just code maintenance in this scenario

HongNguyen635: I was thinking if the key failed, there is no reason to init Bluetooth beforehand. But yes I can make that change to be consistent.

Micahscope: Sound reasoning. The user will still see `hubble_init()` fail if the key failed, so it will be fine either way.

Thanks for taking the time to make it consistent. I think there is good value in trying to make our sample apps look as similar as possible to each other

Micahscope (approved):

ceolin (approved):

PR #329, opened 2026-09-01
crypto: Support referencing the Hubble key by a PSA key id

ceolin:
Add CONFIG_HUBBLE_NETWORK_CRYPTO_PSA_USE_KEY_ID, so the SDK signs through a key
the PSA implementation already holds and never sees the key material.

Includes tools/ncs-generate-key.py to build the nrfutil keyslot JSON, and a ble-beacon
sample setup provisioning the key into CRACEN KMU on nRF54L15

### DIFF
--- a/include/hubble/port/crypto.h
+++ b/include/hubble/port/crypto.h
@@ -102,6 +102,32 @@ int hubble_crypto_cmac(const uint8_t key[CONFIG_HUBBLE_KEY_SIZE],
 		       uint8_t output[HUBBLE_AES_BLOCK_SIZE]);
 
 /**
+ * @brief Computes the Cipher-based Message Authentication Code (CMAC)
+ *        using a key identifier.
+ *
+ * Same as hubble_crypto_cmac(), but the key is referenced by an opaque
+ * key identifier instead of by its key material. The SDK never has
+ * access to the key material: the application provisions the key into a
+ * secure key storage and the CMAC is computed by the crypto provider on
+ * the key it holds internally, so the raw key never leaves that storage.
+ *
+ * @param key_id Buffer holding the identifier of the key used for the
+ *               CMAC calculation, in the format expected by the crypto
+ *               provider (for PSA Crypto, a psa_key_id_t). It does not
+ *               contain key material.
+ * @param data Pointer to the input data for which the CMAC is
+ *             calculated.
+ * @param len The length of the input data in bytes.
+ * @param output Pointer to the buffer where the CMAC (message
+ *               authentication code) will be stored. This buffer must
+ *               be large enough to hold the CMAC value
+ *               (typically the size of the AES block, 16 bytes).
+ * @return 0 on success, non-zero on error.
+ */
+int hubble_crypto_key_id_cmac(const void *key_id, const uint8_t *data,
+			      size_t len, uint8_t output[HUBBLE_AES_BLOCK_SIZE]);
+
+/**
  * @}
  */ /* hubble_crypto */
 
--- a/port/zephyr/Kconfig
+++ b/port/zephyr/Kconfig
@@ -194,6 +194,21 @@ config HUBBLE_NETWORK_CRYPTO_PSA
 	  When building with TF-M, cryptographic operations are
 	  delegated to the secure processing environment.
 
+config HUBBLE_NETWORK_CRYPTO_PSA_USE_KEY_ID
+	bool "Use PSA Crypto opaque key id"
+	depends on HUBBLE_NETWORK_CRYPTO_PSA
+	help
+	  Reference the Hubble key by its PSA opaque key
+	  identifier instead of by its key material.
+
+	  With this option the SDK never needs nor has access to
+	  the Hubble key material. The application provisions the
+	  key into a secure key storage (for example TF-M ITS or
+	  a KMU) and hands the SDK only the key id returned by
+	  that storage. All cryptographic operations are then
+	  performed by the PSA implementation on the key it holds
+	  internally, so the Hubble key material is never directly used.
+
 config HUBBLE_NETWORK_CRYPTO_CUSTOM
 	bool "Custom crypto implementation"
 	help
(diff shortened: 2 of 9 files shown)


port/zephyr/Kconfig
```diff
@@ -194,6 +194,22 @@ config HUBBLE_NETWORK_CRYPTO_PSA
 	  When building with TF-M, cryptographic operations are
 	  delegated to the secure processing environment.
 
+config HUBBLE_NETWORK_CRYPTO_PSA_USE_KEY_ID
+	bool "Use PSA Crypto opaque key id"
+	depends on HUBBLE_NETWORK_CRYPTO_PSA
+	help
+	  eference the Hubble key by its PSA opaque key
```
Micahscope: minor typo - missing r in reference

src/hubble_crypto.c
```diff
@@ -142,8 +142,8 @@ int hubble_internal_sequence_acquire(uint32_t time_counter, uint16_t *seq_no)
 	return ret;
 }
 
-static int _kbkdf_counter(const uint8_t *key, const char *label,
-			  size_t label_len, const uint8_t *context,
+static int _kbkdf_counter(const uint8_t *key, const char *label, size_t label_len,
+			  bool hubble_key, const uint8_t *context,
```
Micahscope, line 146: I usually don't like passing bool as parameters because it isn't clear in the function call what it means.

```
err = _kbkdf_counter(master_key, "DeviceKey",
				     strlen("DeviceKey"), true, context,
				     strlen((const char *)context), output_key,
				     CONFIG_HUBBLE_KEY_SIZE);
```

I like creating an enum and passing in an enum value to ease readability. Something like

```
typedef enum {
  KEY_USE_ID,
  KEY_USE_MATERIAL,

  KEY_USE_COUNT
} key_use_t;
```

Then passing in `KEY_USE_ID` or `KEY_USE_MATERIAL` to the function.

Thoughts on this? Since this function already takes a handful of parameters, this could be helpful in a reader understanding what the function call is doing

ceolin: I agree in general, specially for public APIs. The thing here is creating a new alias for such small scope and a binary option. Would make it clear doing 's/hubble_key/use_key_id/' ?

Micahscope: Yeah that seems clear to me. The one thought I'd have with strings, is that this could start to increase our flash space by comparing to strings that have to be stored in memory. I suppose the same concern could be raised about enums (likely compiled to uint32_t)

Just some thoughts. As long as we consider the tradeoffs, happy to go any direction

ceolin: Compilers are smart and use smaller types, the thing for me is really the scope, having an alias for something that is self content and binary option does not bring much benefits, that is said if we get more type of keys or it needs to be exposed I think your suggestion is the right direction.

Micahscope: Yeah fair. Maybe since this is not a public API, we stay with bool

src/crypto/psa.c
```diff
@@ -43,13 +43,52 @@ static int _psa_status_to_errno(psa_status_t status)
...
 
 	return ret;
 }
 
+int hubble_crypto_key_id_cmac(const void *key, const uint8_t *input,
+			      size_t input_len,
+			      uint8_t output[HUBBLE_AES_BLOCK_SIZE])
+{
+	psa_status_t status;
+	psa_key_id_t key_id = *(psa_key_id_t *)key;
```
HongNguyen635: should be `NULL` check `key` and other input before this?

src/crypto/psa.c
```diff
@@ -43,13 +43,52 @@ static int _psa_status_to_errno(psa_status_t status)
...
 	default:
 		break;
 	}
 
 	return ret;
 }
 
+int hubble_crypto_key_id_cmac(const void *key, const uint8_t *input,
+			      size_t input_len,
+			      uint8_t output[HUBBLE_AES_BLOCK_SIZE])
```
HongNguyen635: should the header and implementation signature matches? The one in the header is:
```
int hubble_crypto_key_id_cmac(const void *key_id, const uint8_t *data,
			      size_t len, uint8_t output[HUBBLE_AES_BLOCK_SIZE]);
```
This is legal in C but would be difficult when troubleshooting when trying to find the header with its implementation.

include/hubble/port/crypto.h
```diff
@@ -101,6 +101,32 @@ int hubble_crypto_cmac(const uint8_t key[CONFIG_HUBBLE_KEY_SIZE],
...
+ *               provider (for PSA Crypto, a psa_key_id_t). It does not
+ *               contain key material.
+ * @param data Pointer to the input data for which the CMAC is
+ *             calculated.
+ * @param len The length of the input data in bytes.
+ * @param output Pointer to the buffer where the CMAC (message
+ *               authentication code) will be stored. This buffer must
+ *               be large enough to hold the CMAC value
+ *               (typically the size of the AES block, 16 bytes).
+ * * @return 0 on success, non-zero on error.
```
HongNguyen635: nit: remove the extra `*`

src/crypto/psa.c
ceolin: sure, I can change it.

src/crypto/psa.c
ceolin: This has to be checked in the common code, it does not make sense for the current implementation as well.

include/hubble/port/crypto.h
ceolin: fixed

src/crypto/psa.c
```diff
@@ -43,13 +43,51 @@ static int _psa_status_to_errno(psa_status_t status)
...
+		ret = -EPERM;
+		break;
 	default:
 		break;
 	}
 
 	return ret;
 }
 
+int hubble_crypto_key_id_cmac(const void *key, const uint8_t *data, size_t len,
```
HongNguyen635: `const void *key_id`?

port/zephyr/Kconfig
```diff
@@ -194,6 +194,21 @@ config HUBBLE_NETWORK_CRYPTO_PSA
...
```
HongNguyen635: Missing `R` in reference

HongNguyen635 (commented):
sorry for being pedantic, just 2 minor points, everything else is looking good.

HongNguyen635 (approved):

PR #331, opened 2026-09-09
sat: boards/espressif: add sat support for ESP32-S31

HongNguyen635:
- Refactor the Espressif sat port to be able to support multiple targets.
- Add support for ESP32-S31.


port/esp-idf/hubblenetwork-sdk/boards/espressif/esp_sat_radio_config.h
```diff
@@ -0,0 +1,39 @@
...
+#define PORT_ESP_SAT_RADIO_CONFIG_H
+
+/* Compensate for on/off delay */
+#if defined(CONFIG_IDF_TARGET_ESP32C6)
+#define ESP_RADIO_OFF_DELAY_US 450U
+#define ESP_RADIO_ON_DELAY_US  70U
+
+#elif defined(CONFIG_IDF_TARGET_ESP32S31)
+#define ESP_RADIO_OFF_DELAY_US 670U
+#define ESP_RADIO_ON_DELAY_US  20U
```
ceolin, line 17: where does numbers come from ? It is worth to add a note about it.

ceolin (dismissed):
Changes look good to me. Minor comment regarding some missing info.

port/esp-idf/hubblenetwork-sdk/boards/espressif/esp_sat_radio_config.h
HongNguyen635: These are just numbers from experiment since I don't have visibility into the blobs, but good point, I'll add a note here.

PR #332, opened 2026-09-11
sat: ti: add comment regarding keep FS on after transmit

HongNguyen635:
Add comment explaining why currently we need to keep FS on after every transmit to avoid performance issue.


port/freertos/boards/ti/cc23xx_cc27xx/radio.c
```diff
@@ -168,7 +168,13 @@ static inline int _board_init(void)
 	rclPacketTxCmdGenericTxTest_ble_custom.common.status =
 		RCL_CommandStatus_Idle;
 
-	/* No whitening, on repeated word, freq synth off after cmd, enable cw */
+	/*
+	 * No whitening, on repeated word, freq synth on after cmd, enable cw.
+	 *
+	 * TODO: turn off fs after tx (per called to enable/disable, not per
```
ceolin: Shouldn't be a warning instead of TODO ?

HongNguyen635: @ceolin I put TODO so I can find it later :P
This is a todo item because it needs a solution, or a trade off decision between current consumption vs performance. But I can make it a warning if that fit the context better.

ceolin: If you keep it as TODO, add the work that has to be done :)
When I read it I just see a notification without further work.

HongNguyen635: okay fixed 🫡

ceolin (approved):

port/freertos/boards/ti/cc23xx_cc27xx/radio.c
ceolin: Thank you :)

PR #333, opened 2026-09-11
crypto: Match signature and implementation parameters name

ceolin:
Matching the parameters name between declaration and implementation.


HongNguyen635 (commented):
can you also change the `hubble_crypto_cmac` to match in all of the TI samples please?

HongNguyen635 (approved):

PR #339, opened 2026-09-17
port: sisdk: add Hubble Device SDK as SiLabs extension

HongNguyen635:
Overview:
- Adds the necessary files to make the Hubble SDK works as a SiSDK extension.
- Add a basic `ble-beacon` sample.

Currently, the project works with Simplicity Studio 6 and the cli tool (`slt`, `slc-cli`). Documentation for quick start, guide, and CI will be added in future PR.


port/sisdk/hubble_config.h
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*
+ * Simplicity SDK configuration for the Hubble Device SDK
+ * expressed as CMSIS Configuration Wizard annotations.
+ */
+
+#ifndef INCLUDE_PORT_SISDK_CONFIG_H
```
Micahscope, line 12: I know you copied the existing naming convention, but I am thinking we should include "HUBBLE" in the include guard names.

It seems unlikely, but not outlandish, that someone developing with SiSDK pulls in Hubble extension, which defines `INCLUDE_PORT_SISDK_CONFIG_H`, and some other 3rd party extension, which also defines this exact generic-ish name for a port header, causing a header to not be included correctly

port/sisdk/hubble_config.h
```diff
@@ -0,0 +1,120 @@
...
+// <i> testing where controlled time values are needed to verify
+// <i> time-dependent behavior.
+// <d> 0
+#define CONFIG_HUBBLE_UPTIME_CUSTOM                        0
+
+// </h>
+
+// <<< end of configuration section >>>
+
+/*
```
ceolin, line 83: I don't like these block bellow very much. I would rather have symbols used defined and options not selected can be comments if you want.

port/sisdk/hubble_config.h
```diff
@@ -0,0 +1,120 @@
...
+#if !CONFIG_HUBBLE_NETWORK_SEQUENCE_NONCE_CUSTOM
+#undef CONFIG_HUBBLE_NETWORK_SEQUENCE_NONCE_CUSTOM
+#endif
+
+#if !CONFIG_HUBBLE_UPTIME_CUSTOM
+#undef CONFIG_HUBBLE_UPTIME_CUSTOM
+#endif
+
+/* Matching the counter source selection to the right config symbol */
+#if CONFIG_HUBBLE_COUNTER_SOURCE
```
ceolin, line 100: This is does not look correct, we have it to group possible options in Kconfig, here you have make an implicit association `CONFIG_HUBBLE_COUNTER_SOURCE == 1`  == `CONFIG_HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME`.

samples/freertos/silabs/ble-beacon/ble_beacon.slcp
```diff
@@ -0,0 +1,68 @@
...
+  - name: configTIMER_TASK_STACK_DEPTH
+    value: "240"
+  - name: SL_PSA_KEY_USER_SLOT_COUNT
+    value: "0"
+    condition:
+      - psa_crypto
+  - name: SL_BOARD_ENABLE_VCOM
+    value: "1"
+  - name: SL_IOSTREAM_EUSART_VCOM_CONVERT_BY_DEFAULT_LF_TO_CRLF
+    value: "1"
```
ceolin, line 65: Shouldn't the SDK options be defined like this instead of using the header ?

samples/freertos/silabs/ble-beacon/src/app.c
```diff
@@ -0,0 +1,179 @@
...
+	vTaskNotifyGiveFromISR(_adv_refresh_task_handle, &higher_prio_task_woken);
+	portYIELD_FROM_ISR(higher_prio_task_woken);
+}
+
+/* Bluetooth stack event handler */
+void sl_bt_on_event(sl_bt_msg_t *evt)
+{
+	sl_status_t status;
+
+	switch (SL_BT_MSG_ID(evt->header)) {
```
ceolin: There is only one branch, it can be just an:
```
if (SL_BT_MSG_ID(evt->header) != sl_bt_evt_system_boot_id) {
    return;
}
...
```

samples/freertos/silabs/ble-beacon/src/app.c
```diff
@@ -0,0 +1,179 @@
...
+
+	default:
+		break;
+	}
+}
+
+static void _adv_refresh_task(void *param)
+{
+	(void)param;
+	sl_status_t status;
```
ceolin: this is used only inside the `for` block

port/sisdk/hubble_sisdk.c
```diff
@@ -0,0 +1,43 @@
...
+		[HUBBLE_LOG_INFO] = APP_LOG_LEVEL_INFO,
+		[HUBBLE_LOG_WARNING] = APP_LOG_LEVEL_WARNING,
+		[HUBBLE_LOG_ERROR] = APP_LOG_LEVEL_ERROR,
+	};
+
+	va_list args;
+	sl_status_t status;
+
+	if (!app_log_check_level(app_level[level])) {
+		return 0;
```
ceolin, line 29: Shouldn't it return `-EINVAL` ?

ceolin (changes requested):
Few comments to review, it is so simple, that is awesome :)

port/sisdk/hubble_config.h
HongNguyen635: adressed this offline, but will put this here in case we need to reference it later. From Flavio: "We can change it, but that is not a big deal. The header is not included by the application which means it will be a problem only if we include a header that defines this before including it. We can address it in a following commit".

samples/freertos/silabs/ble-beacon/src/app.c
HongNguyen635: good point. There were other events but not relevant so I only handled this case and didn't notice that it's only 1 event. I'll update it.

samples/freertos/silabs/ble-beacon/src/app.c
HongNguyen635: I have the habit of define everything at the top of the function. Will move it.

port/sisdk/hubble_sisdk.c
HongNguyen635: This is not an error though, if your level is only upto info, and the caller is debug type, it should be supressed. The [zephyr port](https://github.com/HubbleNetwork/hubble-device-sdk/blob/799d3adadeea1cf85e563e101e0989e25be99552/port/zephyr/hubble_zephyr.c#L40) also do something similar.

port/sisdk/hubble_config.h
HongNguyen635: I also thought about just reuse the `config.h` header from FreeRTOS port, but this is the style that SiLabs use, which will show up in the Configurator UI like so:

image


Whereas the plain `#define` block doesn't show up. An example reference I used is [this file from SiSDK](https://github.com/SiliconLabs/simplicity_sdk/blob/sisdk-2025.6/app/bluetooth/common/ots/config/sl_bt_ots_server_inst_config.h).

samples/freertos/silabs/ble-beacon/ble_beacon.slcp
HongNguyen635: This way allows you to define arbitrary configs or override existing one, while having the header ties it to the SDK extension, so it can show up in the UI and have default values. We can remove the header, but then user have to manually copy/add these options on every new project.

port/sisdk/hubble_config.h
HongNguyen635: I spent some time trying to find a workaround for this. But this is the best I could come up with.
I can make it like this:
```
// <o CONFIG_HUBBLE_COUNTER_SOURCE> Counter source
// <HUBBLE_COUNTER_SOURCE_UNIX_TIME=> Unix time
// <HUBBLE_COUNTER_SOURCE_DEVICE_UPTIME=> Device uptime
#define CONFIG_HUBBLE_COUNTER_SOURCE HUBBLE_COUNTER_SOURCE_UNIX_TIME
```
but I can't compare `#if CONFIG_HUBBLE_COUNTER_SOURCE == HUBBLE_COUNTER_SOURCE_UNIX_TIME` unless I assigned value to it, which goes back to the implicit value association.
I can name the symbol different than `CONFIG_HUBBLE_COUNTER_SOURCE` to prevent the confusion, but that is the best way I could think of.

